Issue #5, opened 2016-02-21
Off by one error in division


AB95:
Nim:
![image](https://cloud.githubusercontent.com/assets/10396224/13204556/f385ec8c-d8ca-11e5-9671-1ce530ea3ff7.png)

Wolfram Alpha:
![image](https://cloud.githubusercontent.com/assets/10396224/13204559/fa3cba9c-d8ca-11e5-9505-7d108910253b.png)

Python:
![image](https://cloud.githubusercontent.com/assets/10396224/13204656/4ebccbc8-d8cd-11e5-9ed0-6a0aa26d9750.png)

Nim v0.13.0
Bigints downloaded earlier today, so also up to date

The calculation is 40!/(20!*20!)

dom96:
Next time please don't put the code in an image.

```nim
import bigints

var x = initBigInt("815915283247897734345611269596115894272000000000")
var y = initBigInt("5919012181389927685417441689600000000")

echo(x div y)
```

Interesting bug though.

@def- Any ideas?

def-:
This part of the division gets called and probably causes the off-by-one, but I'm not sure yet where the mistake is: https://github.com/def-/nim-bigints/blob/master/src/bigints.nim#L609-L616

AB95:
Discovered a problem with this as well:

```nim
import bigints

var x = initBigInt("815915283247897734345611269596115894272000000000")
var y = initBigInt("20000000000")

echo x div y
```

It's returning 40795764162394886701185424564904152901 instead of 40795764162394886717280563479805794713. A difference of 16095138914901641812. 

As this code goes through the branch you mentioned there may be an issue with, I thought it could be the same problem.

PR #6, opened 2016-06-07
Changed division to handle certain edge cases

AB95:
Zhi could be off in rare edge cases. In the subtraction phase, q.limbs[v+i] + r.limbs[i] returned a uint32 before being added to carry, which could result in overflow.


def-:
Nice, you found it. Closes #5!

PR #7, opened 2016-06-10
Implemented initBigInt for int64 and uint64

AB95:
Had some issues when trying to include them in the existing method, but this seems to be working.


def-:
Looks good, thanks.

PR #8, opened 2016-06-16
Add overloaded  operators for BigInt types

msmith491:
Just making it easier to make assumptions about bigint library behavior based on normal Nim behavior

### DIFF
--- a/src/bigints.nim
+++ b/src/bigints.nim
@@ -742,6 +742,10 @@ proc `divmod` *(a, b: BigInt): tuple[q, r: BigInt] =
 #  var tmp = null
 #  division(tmp, x, y, z)
 
+proc `%%` *(a: BigInt, b: int32): BigInt = a mod b
+
+proc `%%` *(a, b: BigInt): BigInt = a mod b
+
 template optDivMod*{w = y div z; x = y mod z}(w,x,y,z: BigInt) =
   division(w, x, y, z)
 
--- a/tests/tester.nim
+++ b/tests/tester.nim
@@ -60,3 +60,20 @@ test "off by one in division (https://github.com/def-/nim-bigints/issues/5)":
     var x = initBigInt("815915283247897734345611269596115894272000000000")
     var y = initBigInt("20000000000")
     check x div y == initBigInt("40795764162394886717280563479805794713")
+
+test "Mod operators":
+  block:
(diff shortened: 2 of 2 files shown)


def-:
Careful, I believe `%%` implements unsigned modulo and is not just an overload: http://nim-lang.org/docs/system.html#%%,int64,int64

msmith491:
I'll take another look at it and make sure I handle the correct case for the operator

msmith491:
So, the implementation of `division` in num-bigints appears to be unsigned by default:
https://github.com/def-/nim-bigints/blob/master/src/bigints.nim#L689

So while the original might not just be a straight overload, this one might be since the implementation of `mod` is unsigned in the bigints.nim module.

def-:
Well, then division should rather be fixed! (Note the TODO there)

msmith491:
On an semi-related note, how do you get the nim compiler to use the local copy of a library instead of the one installed via nimble?

msmith491:
I think I came into this with the assumption that the modulo operator worked similarly to how it does in Python, the result of the operation has the same sign as the divisor:

Eg:  
In python 10 mod 3 == 1
In nim      10 mod 3 == 1

In python 10 mod -3 == -2
In nim      10 mod -3 == 1

The difference appears to be that nim follows the C/C++ philosophy that the sign follows the divisor instead of the dividend.  It gets weird when I try to do comparisons between `%%` and `mod`.

All positive numbers are fine
10 %% 3 == 1
10 mod 3 == 1

But when we get to negatives, all hell breaks loose
10 %% -3 == 10  ??? What the hey?
10 mod -3 == 1   <-- This one makes sense

-10 %% 3 == 0
-10 mod 3 == -1

-10 %% -3 == -10  
-10 mod -3 == -1

It's like `mod` is treating everything as unsigned, then having the result get the sign of the dividend
`%%` is just craziness that I'm having trouble interpreting.

This has been educational, and I'll probably just avoid `%%` indefinitely

def-:
> 10 %% -3 == 10 ??? What the hey?
Assuming you're on 64bit, -3 is 18446744073709551613, so you're calculating `10 mod 18446744073709551613` which is obviously 10.

msmith491:
Oh...that makes more sense.

PR #9, opened 2016-06-16
Add ability to call initBigInt on an existing BigInt

msmith491:
Makes it easier to write generic code if we can get an initBigInt operation that does nothing on an existing BigInt


def-:
That's a bit like calling `int()` converter on an int, so I guess it's fine.

Issue #16, opened 2018-01-24
Zero result from subtraction always has Negative flags


Sud0nim:
Hi Dennis,

Forgive me if I have misunderstood a reason for this, but I noticed when using subtraction to zero (or addition on equal magnitude but opposite signed BigInts), the resulting zero value always has Negative flags. EDIT: I added an example using multiplication, as I also noticed the same behaviour.

My expectation was that a zero value would normally be given as a positive zero by default, as that is typically how I would expect a native integer type to behave. 

The negative flag can be worked around if you are aware of it, but I am not sure it is obvious or expected behaviour for the end user.

For example:

```nim
var 
  a = initBigInt("828478292990482")
  b = initBigInt(9283)
  c = initBigInt("-828478292990482")
  d = initBigInt(-9283)
  e = initBigInt(0)


echo b + d
echo a + c

echo d + b
echo c + a

echo b - b
echo d - d

echo d * e

# Outputs:

-0
-0
-0
-0
-0
-0
-0
```

Is there any reason for the result to always be given a negative sign, or are you happy for this to default to positive (un-flagged) zero?

Thank you

def-:
Sounds like a bug, should be positive.

Sud0nim:
OK sure, thanks.

Do you want me to submit a pull request? 

My solution may not be optimal, but would be something to the effect of adding a final check to each of the arithmetic operators as a temporary measure:

```nim
proc unsignedSubtraction(a: var BigInt, b, c: BigInt) =
  if unsignedCmp(b, c) > 0:
    realUnsignedSubtraction(a, b, c)
  else:
    realUnsignedSubtraction(a, c, b)
    negate(a)
  if a.limbs == @[0'u32]: # Add a check to see if the resulting BigInt has a magnitude of zero
    a.flags.excl(Negative)
```
Which passes the existing tester.nim and my own quick set of tests.

However, you may have a more elegant or efficient solution of course.

def-:
Sounds good to me, you can make a PR of course, thank you.

PR #17, opened 2018-01-25
Remove negative flags for zero values

Sud0nim:
Hi Dennis,

This should resolve [#16](https://github.com/def-/nim-bigints/issues/16)

I also added some tests for this particular issue.

Thanks


def-:
Thanks

PR #19, opened 2019-03-19
fix compilation on latest devel

SolitudeSF:
i missed what happened to ranges, but it seems like we need to explicitly convert them.


def-:
Merci!

PR #28, opened 2020-07-02
Export Flags.Negative

ehmry:
Export this enum to construct a negative Bigint instance in-place.

### DIFF
--- a/src/bigints.nim
+++ b/src/bigints.nim
@@ -1,7 +1,7 @@
 import strutils
 
 type
-  Flags = enum
+  Flags* = enum
     Negative
 
   BigInt* = tuple


pietroppeter:
would exporting the enum be enough or would you also need to export `flags` field?
Another option would be to export the `negate` proc.

ehmry:
> would you also need to export `flags` field?

No, BigInt is a tuple so the fields are implicitly exported.

> Another option would be to export the `negate` proc.

Yes, but only if there is some sugar to check if a `BigInt` is negative.

ehmry:
FWIW I'm adding bignum support to the CBOR library, and hackishly pulling limbs from the stream: https://git.sr.ht/~ehmry/nim_cbor/tree/4519112bddeebe624134a5a0d2b14339a0e99e2c/src/cbor/bignum.nim.

PR #31, opened 2020-10-04
Add api docs in readme

pietroppeter:
Hi, I added API documentation to the readme using https://github.com/treeform/mddoc (I have to file a PR there too to fix an issue with proc using '<' or '>').

It is not a big improvement over watching source directly (basically it only lists public interface, but still it is better than nothing and I do find it already useful.

I actually plan to keep coming and improving a bit more documentation and hopefully also start addressing the other open issues.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -106,42 +106,42 @@ proc cmp(a: int32; b: BigInt): int64
 proc cmp(a: BigInt; b: int32): int64
 ```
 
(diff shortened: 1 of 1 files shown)


def-:
> I actually plan to keep coming and improving a bit more documentation and hopefully also start addressing the other open issues.

Great, glad to hear.

Issue #26, opened 2019-12-22
Empty limbs when uninitialized


auxym:
Consider the following:

```nim
var
    a: BigInt
    b: BigInt = 12.initBigInt

echo a*b
```
This results in the following stack trace:

```
[...]/scratch.nim(7) scratch
[...].nimble/pkgs/bigints-0.4.3/bigints.nim(484) *
[...].nimble/pkgs/bigints-0.4.3/bigints.nim(394) multiplication
[...].choosenim/toolchains/nim-#devel/lib/system/fatal.nim(39) sysFatal
Error: unhandled exception: index out of bounds, the container is empty [IndexError]
```

This is because `a.limbs` is `@[]`. Would it be possible for an uninitialized `BigInt` to default to `limbs = @[0]`, to be consistent with nim integers? Otherwise, maybe there could be a flag `initialized` that is checked before trying to operate on a `BigInt`, so we get slightly more readable errors. I'm not super proud about it, but tbh this took me a while to debug. To be fair, my actual code was slightly more complex and resulted in a zero BigInt not being initialized in a corner case.

def-:
Sounds good but would require https://github.com/nim-lang/RFCs/issues/126

auxym:
Yeah, I wasn't sure if/how it could be implemented in Nim. Good to know there's an RFC about it!

PR #32, opened 2020-10-06
Improve documentation

pietroppeter:
Hi,
another small incremental improvement for documentation:

- added a short summary of package features
- added a list of current limitations and potential enhancements (taken from TODO list and issues)

There is one element of TODO list that I think it should not be an issue anymore (since there are no more nil for sequences) and so I did not list it in current limitations: "Fix the templates so they work with uninitialized (nil) values!"

Let me know what you think.

Regarding possible API changes in the future I have always wondered why Negative is an enum instead of a bool (where there other flags at the beginning?), but for the moment this is just curiosity, plan is still to fix outstanding issues before thinking of the improvements. I am enjoying the process and learning quite a bit, the code is very terse.

In the meantime I have started to work on #26 (see https://github.com/pietroppeter/nim-bigints/tree/fix-empty-limbs). I think there are many places where empty limbs might raise an issue, the plan is map them feature by feature with failing tests and fix them one by one. I will open a draft PR when I complete fixes for the compare operators and proceed from there.

Also notable is a new bigints library that was born a couple of days ago and which probably aims to have same footprint as bigints (I do expect it to progress fast and be very performant given its author): https://github.com/SciNim/megalo/tree/master/megalo

### DIFF
--- a/TODO
+++ b/TODO
@@ -1,1 +1,0 @@
-- Fix the templates so they work with uninitialized (nil) values!
--- a/TODO
+++ b/TODO
@@ -1,8 +1,1 @@
-- Can't multiply number with itself x *= x, probably also same for other operations
-- Should probably use ref object instead of tuple for BigInt type
-- Fix the library on 32bit
 - Fix the templates so they work with uninitialized (nil) values!
-- karatsuba multiplication
-- better addition
-- everything in combination with normal ints and uints too
-(- better division)
--- a/readme.md
+++ b/readme.md
@@ -26,6 +26,16 @@ Most of the operations above (all those for which it makes sense) are also avail
 
 For examples of usage see the [examples](examples) folder.
 
+## Current limitations and possible enhancements
+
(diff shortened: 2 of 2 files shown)


def-:
Hm, so are you still planning to work on this? I think it would make more sense to combine the efforts in one project, especially considering NIm's community size. mratsim probably had good reasons to rewrite instead of adapting my code, so I'd talk to him if I were you. If megalo will be better than nim-bigints and a drop-in-replacement, we could add a big link to it in the readme.

pietroppeter:
well, yeah, for the moment I still plan to work on bigints. It is a good and simple library and it seems to me it just needs a bit of effort to be more "finished". I agree that it will make sense eventually to discuss and possibly join forces, but for the moment bigints is here and megalo has just started (and I guess it is probably more difficult to contribute to it in this initial phase). Maybe this will change in a week or so. I do think the effort I put in bigints right now (which is mostly learning on my side), will be still useful to possibly contributing to megalo in the future or discuss with mratsim on what are the best options for this two libraries and for optimizing effort.

PR #33, opened 2020-10-06
Fix empty limbs

### LINKED ISSUE #26: Empty limbs when uninitialized

pietroppeter:
Fixes #26 handling @[] the same as @[0]

- [X] comparison
- [X] addition/subtraction
- [X] multiplication
- [X] others place where empty limbs can cause issues? did not find any. division looks fine
- [X] bump version
- [X] update readme (rebase on current master)
- [X] cleanup
- [X] rerun mddoc (since public interface change, see below)

other change in this PR: `one` and `zero` are now public.

### DIFF
--- a/bigints.nimble
+++ b/bigints.nimble
@@ -1,6 +1,6 @@
 # Package
 
-version     = "0.4.3"
+version     = "0.4.4"
 author      = "Dennis Felsing"
 description = "Arbitrary-precision integers implemented in pure Nim"
 license     = "MIT"
--- a/readme.md
+++ b/readme.md
@@ -29,7 +29,6 @@ For examples of usage see the [examples](examples) folder.
 ## Current limitations and possible enhancements
 
 * cannot multiply a number with itself (x *= x). An issue like this exist also for addition (#27) and possibly other operations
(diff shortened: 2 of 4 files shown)


def-:
Nice, thanks!

PR #38, opened 2021-11-05
Fix int and BigInt comparison like 2 < 100.initBigInt

EyeCon:
The following check used to fail: `check 2 < 100.initBigInt`.

I think this comes down to a basic typo in one line. I also added some simple test cases for this.


def-:
Thanks

PR #39, opened 2021-11-10
add disclaimer about the move to nim-lang/bigints

pietroppeter:
Might I add here big thanks 🙏 to @def- for this, we miss you in Nim community! 😊

https://github.com/nim-lang/Nim/issues/14696#issuecomment-962924051


def-:
How about I move my repo to nim-lang/bigints using the github settings instead? Then everyone landing in def-/nim-bigints would automatically be redirected to nim-lang/bigints.

pietroppeter:
> How about I move my repo to nim-lang/bigints using the github settings instead?

ah, yes, I agree it would be a better option (stars would also be inherited), but it would require some additional effort on aligning the repo. nim-lang/bigints was forked some time ago and commit history diverged. I think the effort would be worth though. I will close this PR now (we can always open it again) and we could probably move this discussion on https://github.com/nim-lang/Nim/issues/14696

PR #54, opened 2021-12-09
Fix & optimize `toString`, `initBigInt`

konsumlamm:
I added all bugs that I found (and more) to the test suite. `calcSizes` now produces other results (though only for powers of two, so we don't have to use `sizes[base] + 1` for those anymore) and is more efficient (due to using multiplication rather than division). `initBigInt` is now also optimized for bases which are a power of two (similar to `toString`).

### DIFF
--- a/src/bigints.nim
+++ b/src/bigints.nim
@@ -1,7 +1,7 @@
 ## Arbitrary precision integers.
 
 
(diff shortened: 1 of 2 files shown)


konsumlamm:
@narimiran ping

PR #57, opened 2021-12-31
Improve `pow`, `shl`, `shr`

konsumlamm:
* make the second argument of `pow`, `shl`, `shr` be of type `Natural`, since negative values make no sense here
* use `and`/`shr` instead of `mod`/`div` in `pow` 
* rewrite `shl` to use a similar algorithm to `shr`, rather than multiplying by a power of 2
* add a bunch of tests

EDIT:
* fix `shl`, `shr` to work correctly for negative values


konsumlamm:
I found a (pre-existing) bug in `shr` for negative values, I'll put together a fix for that.

Issue #53, opened 2021-12-08
Turn `proc`s into `func`s


konsumlamm:
As far as I can tell, all `proc`s could be turned into `func`s, providing the guarantee that they don't have side effects. Is there anything or anyone opposing this?

dlesnoff:
Can we close this issue ? Or do we need #62 to be merged before ?

konsumlamm:
#61 already implemented this. For the future: if you link an issue with "fixes #53" or "closes #53" in a PR, it gets automatically closed when the PR is merged.

PR #61, opened 2022-01-01
Change procs into funcs

dlesnoff:
Changed procs into funcs in accordance to standard library solving Issue number #53.


konsumlamm (approved):

xflywind (approved):
tests can probably enable the strictFuncs fearure.
ref https://nim-lang.org/docs/manual_experimental.html#strict-funcs

PR #63, opened 2022-01-03
Improve division/modulo code

konsumlamm:
Division/modulo by zero now raises a `DivByZeroDefect`. Before it would fail with a SIGFPE.

### DIFF
--- a/src/bigints.nim
+++ b/src/bigints.nim
@@ -568,11 +568,10 @@ func unsignedDivRem(q: var BigInt, r: var uint32, n: BigInt, d: uint32) =
   q.limbs.setLen(n.limbs.len)
   r = 0
   for i in countdown(n.limbs.high, 0):
-    let tmp: uint64 = uint64(n.limbs[i]) + uint64(r) shl 32
+    let tmp = uint64(n.limbs[i]) + uint64(r) shl 32
     q.limbs[i] = uint32(tmp div d)
     r = uint32(tmp mod d)
-  while q.limbs.len > 1 and q.limbs[q.limbs.high] == 0:
-    q.limbs.setLen(q.limbs.high)
+  normalize(q)
 
 func bits(d: uint32): int =
   const bitLengths = [0, 1, 2, 2, 3, 3, 3, 3, 4, 4, 4, 4, 4, 4, 4, 4,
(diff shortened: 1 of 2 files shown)


src/bigints.nim
```diff
@@ -615,32 +613,32 @@ func unsignedDivRem(q, r: var BigInt, n, d: BigInt) =
...
     a.limbs.setLen(k)
     let wm1 = r.limbs[r.limbs.high]
     let wm2 = r.limbs[r.limbs.high-1]
     var ak = k
 
-    var zhi = 0.initBigInt
-    var z = 0.initBigInt
-    var qib = 0.initBigInt
-    var q1b = 0.initBigInt
+    var zhi = zero
```
dlesnoff, line 622: We can make all these initialisations either in one line :
```Nim
var zhi, z, qib, q1b = zero
```
or with only one var and an indented block and group with preceding initializations of a and ak.

src/bigints.nim
```diff
@@ -696,25 +694,24 @@ func unsignedDivRem(q, r: var BigInt, n, d: BigInt) =
...
   r.isNegative = n < 0 and r != 0
 
   # divrem -> divmod
-  if (r < 0 and d > 0) or
-     (r > 0 and d < 0):
+  if (r < 0 and d > 0) or (r > 0 and d < 0):
     r += d
     q -= one
 
-  if q.limbs == @[0'u32]:
```
dlesnoff, line 710: Why can this code be safely removed ? Do we care the sign of 0 ?
(isZero does not check the sign). It could be interesting to know for the user that +0 = -0.

src/bigints.nim
```diff
@@ -696,25 +694,24 @@ func unsignedDivRem(q, r: var BigInt, n, d: BigInt) =
...
-    q.isNegative = false
-
-  if r.limbs == @[0'u32]:
-    r.isNegative = false
-
```
dlesnoff, line 715: Why can this code be safely removed ? Do we care the sign of 0 ?
(isZero does not check the sign). It could be interesting to know for the user that +0 = -0.

dlesnoff: There is this (old) issue that mentioned a similar problem in subtraction :
  Zero result from subtraction always has Negative flags #16
  Does your other changes prevent a negative zero ?

dlesnoff (approved):

Issue #45, opened 2021-11-25
[Feature request] `powmod`


konsumlamm:
### Description

Provide a `powmod(base, exponent, modulus)` function that computes `(base ^ exponent) % modulus`, but more efficiently than the naive computation. In particular, it should avoid building up `BigInt`s that are substantially larger than `modulus`.

### Related work

* [D](https://dlang.org/phobos/std_bigint.html#.powmod)
* [Rust](https://docs.rs/num-bigint/0.4.3/num_bigint/struct.BigInt.html#method.modpow)

dlesnoff:
We should first implement a gcd function and a bitops operation to count the number of bits.
I tried to implement it for prime modulus and to add some tests. I have to add sign handling too. I am unsure how to handle composite numbers. 
From a mathematical point of view, I would reduce the exponent modulo the euler totient function of the modulus, and compute the power modulo each factor. We compute the result back with CRT (Chinese Remainder Theorem).
But computing the euler totient is as hard as factoring the integer (for very large modulus). So there are probably more practical way. Maybe we should give the euler totient as an optional parameter ? It would avoid check if the modulus is prime and to factor it.
I checked Rust library but don't understand their logic. For BigInt, they call a function of another module named power (so I guess they do it the naive way ?). For BigUint, they use a "Montgomerry Ladder" for odd integers implemented in another file I could not find, and for even integers  I have not check it yet, but is handmade and optimized.

konsumlamm:
For now, we could just implement it like `pow` (using binary exponentiation, but take the modulo every iteration, like the [D implementation](https://github.com/dlang/phobos/blob/4687883231eba3bda7691321f2af107fdb3d0a44/std/bigint.d#L2328) does) and optimize it later.

PR #64, opened 2022-01-04
Add powmod to BigInt

dlesnoff:
Following request of Issue #45, I have implemented modular exponentiation following SageMath implementation. The only discrepancy being it does not yet accept negative exponent.
I have added some tests too. Some checks uses mathematical properties as the little fermat's theorem. We might want to reduce them in the future, as they take a little time.
For negative exponentiation, we need to invert first using an extended euclidean algorithm.
Some examples did not compile anymore, due to a change to the backwards index operator .. < needs to be written as ..< .

### DIFF
--- a/src/bigints.nim
+++ b/src/bigints.nim
@@ -959,3 +959,33 @@ iterator `..<`*(a, b: BigInt): BigInt =
   while res < b:
     yield res
     inc res
+
+func powmod*(base, exponent, modulus: BigInt): BigInt =
+  ## Compute modular exponentation of `base` with power `exponent` modulo `modulus`.
+  if modulus == 0:
+    raise newException(DivByZeroDefect, "modulus must be nonzero.")
+  if modulus == 1:
+    return zero
+  if exponent < 0:
+    # let baseInv = inverse_mod(base, modulus)
+    # return powmod(baseInv, -exponent, modulus)
+    raise newException(ValueError, "Not yet Implemented!")
+  if exponent == 0:
+    if base == 0:
+      raise newException(ValueError, "0^0 is undefined.")
+    return one
+
+  var
+    exp = exponent
+    basePow = ((base mod modulus) + modulus) mod modulus # Base stays in [0, m-1]
+  while (exp and one) == zero:
+    basePow = basePow * basePow mod modulus
+    exp = exp shr 1
+  result = basePow
+  exp = exp shr 1
+  while exp != zero:
+    basePow = (basePow*basePow) mod modulus
+    if (exp and one) != zero:
+      result = (result*basePow) mod modulus
+    exp = exp shr 1
+
--- a/src/bigints.nim
+++ b/src/bigints.nim
@@ -9,14 +9,14 @@ type
     isNegative: bool
 
 
-proc normalize(a: var BigInt) =
+func normalize(a: var BigInt) =
   for i in countdown(a.limbs.high, 0):
     if a.limbs[i] > 0'u32:
       a.limbs.setLen(i+1)
       return
   a.limbs.setLen(1)
 
-proc initBigInt*(vals: sink seq[uint32], isNegative = false): BigInt =
+func initBigInt*(vals: sink seq[uint32], isNegative = false): BigInt =
   ## Initializes a `BigInt` from a sequence of `uint32` values.
   runnableExamples:
     let a = @[10'u32, 2'u32].initBigInt
@@ -25,7 +25,7 @@ proc initBigInt*(vals: sink seq[uint32], isNegative = false): BigInt =
   result.limbs = vals
   result.isNegative = isNegative
 
-proc initBigInt*[T: int8|int16|int32](val: T): BigInt =
+func initBigInt*[T: int8|int16|int32](val: T): BigInt =
   if val < 0:
     result.limbs = @[(not val).uint32 + 1] # manual 2's complement (to avoid overflow)
     result.isNegative = true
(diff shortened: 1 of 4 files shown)


src/bigints.nim
```diff
@@ -959,3 +959,33 @@ iterator `..<`*(a, b: BigInt): BigInt =
   while res < b:
     yield res
     inc res
+
+func powmod*(base, exponent, modulus: BigInt): BigInt =
+  ## Compute modular exponentation of `base` with power `exponent` modulo `modulus`.
+  if modulus == 0:
+    raise newException(DivByZeroDefect, "modulus must be nonzero.")
```
konsumlamm, line 966:
```suggestion
    raise newException(DivByZeroDefect, "modulus must be nonzero")
```
Current error messages don't end with `.` either (besides, this is not a full sentence anyway).

src/bigints.nim
```diff
@@ -959,3 +959,33 @@ iterator `..<`*(a, b: BigInt): BigInt =
...
+  if modulus == 1:
+    return zero
+  if exponent < 0:
+    # let baseInv = inverse_mod(base, modulus)
+    # return powmod(baseInv, -exponent, modulus)
+    raise newException(ValueError, "Not yet Implemented!")
+  if exponent == 0:
+    if base == 0:
+      raise newException(ValueError, "0^0 is undefined.")
```
konsumlamm, line 975: For `pow` and ``math.`^` ``, we have 0^0 = 1, which is also common in practice (although of course it's technically indeterminate). I'm not sure how common this is in modular arithmetic though.

src/bigints.nim
```diff
@@ -959,3 +959,33 @@ iterator `..<`*(a, b: BigInt): BigInt =
...
+    return one
+
+  var
+    exp = exponent
+    basePow = ((base mod modulus) + modulus) mod modulus # Base stays in [0, m-1]
```
konsumlamm, line 980: So the final result will always be in [0, m-1]? What happens if `modulus` is negative? This should be documented.

src/bigints.nim
```diff
@@ -959,3 +959,33 @@ iterator `..<`*(a, b: BigInt): BigInt =
...
+  while (exp and one) == zero:
+    basePow = basePow * basePow mod modulus
+    exp = exp shr 1
+  result = basePow
+  exp = exp shr 1
```
konsumlamm, line 985: Why is this part needed? Is it just an optimization?

src/bigints.nim
```diff
@@ -959,3 +959,33 @@ iterator `..<`*(a, b: BigInt): BigInt =
...
+  while exp != zero:
+    basePow = (basePow*basePow) mod modulus
+    if (exp and one) != zero:
+      result = (result*basePow) mod modulus
```
konsumlamm, line 989:
```suggestion
    basePow = (basePow * basePow) mod modulus
    if (exp and one) != zero:
      result = (result * basePow) mod modulus
```

tests/tbigints.nim
```diff
@@ -379,5 +379,37 @@ template main() =
     doAssert pow(zero, 0) == one
     doAssert pow(zero, 1) == zero
 
+  block: #pow_mod
```
konsumlamm, line 382:
```suggestion
  block: # powmod
```

tests/tbigints.nim
```diff
@@ -379,5 +379,37 @@ template main() =
...
+    let a = "30292868".initBigInt
+    let p = "60449131".initBigInt # p is prime
+    let two = 2.initBigInt
+    doAssert powmod(a, two, p) == "25760702".initBigInt
+    # Fermat's little theorem: a^p ≡ a, for all a in F_p
```
konsumlamm, line 387:
```suggestion
    # Fermat's little theorem: a^p ≡ a mod p
```
I think this is clearer for people that don't know what `F_p` is.

tests/tbigints.nim
```diff
@@ -379,5 +379,37 @@ template main() =
...
+    doAssert powmod(a, p, p) == a
+    # Euler's identity a^(p-1) \equiv 1 \bmod p
+    doAssert powmod(a, p - one, p) == one
+    # We can invert a using euler's identity / fermat's little theorem
```
konsumlamm, line 391:
```suggestion
    # We can invert a using Euler's identity / Fermat's little theorem
```

tests/tbigints.nim
```diff
@@ -379,5 +379,37 @@ template main() =
...
+    doAssert powmod(a, p - two, p) == "51713091".initBigInt
+    # We can reduce the exponent modulo phi(p) = p - 1, since p is prime
+    doAssert powmod(a, 2.initBigInt*p, p) == (a*a mod p)
+
+    let p2 = 761.initBigInt
+    var a2 = 1.initBigInt
+    # Fermat's little theorem: a^p ≡ a, for all a in F_p
```
konsumlamm, line 398:
```suggestion
    # Fermat's little theorem: a^p ≡ a mod p
```
See above.

konsumlamm:
Can you rebase on the latest master? That also makes the diff easier to read.

src/bigints.nim
dlesnoff: If this is true for other functions, then it looks like it would be nitpicking to raise a ValueError for this special case. I have removed it.

src/bigints.nim
dlesnoff: The final result will always be in [0, m-1] contrary to the `mod` keyword for integers.
It does not make sense for the modulus to be negative, I will add an exception.
I have not tested what happens if the base is negative. I will add some tests for this case.

src/bigints.nim
dlesnoff: It does not work without, I am not sure why, I will look into it tomorrow.

dlesnoff:
I have deleted my fork and started from anew, as I do not know how to rebase upstream repositories. I will probably have to make another pull request.

src/bigints.nim
konsumlamm: I think it should work if you replace this by `result = one`. Then we shouldn't need an extra case for `exponent == 0` either.

PR #65, opened 2022-01-05
Add modular exponentiation and modular inverse, closes #45

### LINKED ISSUE #45: [Feature request] `powmod`

dlesnoff:
Add powmod with all the comments made on the previous pull request.


src/bigints.nim
```diff
@@ -959,3 +959,28 @@ iterator `..<`*(a, b: BigInt): BigInt =
   while res < b:
     yield res
     inc res
+
+func powmod*(base, exponent, modulus: BigInt): BigInt =
+  ## Compute modular exponentation of `base` with power `exponent` modulo `modulus`.
+  ## Return value is always in [0, m-1].
```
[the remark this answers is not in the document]
dlesnoff: I am ok for both of these changes. Do I need to rebase ?

src/bigints.nim
```diff
@@ -758,7 +758,7 @@ func `divmod`*(a, b: BigInt): tuple[q, r: BigInt] =
 
 func calcSizes(): array[2..36, int] =
   for i in 2..36:
-    var x = int64(i)
+    var x = i
```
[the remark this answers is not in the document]
dlesnoff: I made a cp of an old file into the repository, this is a git error sorry.

konsumlamm (approved):

dlesnoff:
I finally made a rebase as the changes seemed too minor for a commit. I hope this will not break something, as I had to force push into my branch.

dlesnoff:
I have add modular inverse for negative exponents in modular exponentiation. I fear that modular inverse is too slow. It might be improved with a faster multiplication. Fixes #45.

src/bigints.nim
```diff
@@ -959,3 +959,51 @@ iterator `..<`*(a, b: BigInt): BigInt =
...
+  ## Compute modular exponentation of `base` with power `exponent` modulo `modulus`.
+  ## The return value is always in the range `[0, modulus-1]`.
+  if modulus < 0:
+    raise newException(ValueError, "modulus must be strictly positive")
+  if modulus == 0:
+    raise newException(DivByZeroDefect, "modulus must be nonzero")
+  if modulus == 1:
+    return zero
+  if exponent < 0:
+    if base == zero:
```
pietroppeter, line 994: should we have base == zero return zero also for positive exponent? (and maybe we also check for exponent == 0 before and return 1 in that case)?
code would then be:

```nim
  if exponent == zero:
    return one
  if base == zero:
    return zero
  if exponent < 0:
    let baseInv = invmod(base, modulus)
    return powmod(baseInv, -exponent, modulus)
```

also: not sure if all those ifs would be better with an `if..elif..else` (even if they raise or return)

src/bigints.nim
```diff
@@ -960,6 +960,27 @@ iterator `..<`*(a, b: BigInt): BigInt =
...
+      q, rt, ut : BigInt
+    while r1 > 0:
+      q = r0 div r1
+      rt = r0
+      ut = u
+      r0 = r1
+      u = u1
+      r1 = rt - q * r1
+      u1 = ut - q * u1
+    result = ((u mod modulus) + modulus) mod modulus
```
konsumlamm, line 982: This is only correct if `gcd(a, modulus) == 1`, Otherwise, the modular inverse doesn't exist.

konsumlamm (commented):
I'm not fully done with reviewing, will finish that later (hopefully today).

src/bigints.nim
dlesnoff: Your propositiion removes a lot of unuseful computation for zero base and large positive exponent indeed.
I have not written it this way as zero does not appear as a specific case of the entire algorithm. The square and multiply algorithm would work with 0 as base too.
It is an edge case for modular inverse though and 0 has strictly speaking no modular inverse. As @konsumlamm has proposed to me to remove an if clause in previous PR, I think it is better to keep only edge cases in the first ifs (it improves readability).
Concerning multiple ifs rather than if .. elif .. else clause, I personnaly prefer to avoid else clause as it adds unuseful indentation for the rest of the algorithm. Maybe the compiler is able to do more compilation's optimization with if .. elif .. else clause. I do not think it improves a lot.

src/bigints.nim
```diff
@@ -960,6 +960,27 @@ iterator `..<`*(a, b: BigInt): BigInt =
     yield res
     inc res
 
+func invmod*(a, modulus: BigInt): BigInt =
+    ## Compute the modular inverse of `a` by Euclide's method.
```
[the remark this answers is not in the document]
dlesnoff: I agree and will fix it with a normal comment.

src/bigints.nim
dlesnoff: I can add the test before in the function, but I will need to merge this pull request with the #66 with gcd. Either we merge the gcd PR in the lib before, or I stack them into one branch. The former would make a cleaner history log.

dlesnoff: We actually compute the gcd with the euclid's method, so no need to merge PR's. I can just check that the gcd is one at the end of euclid's computation.

src/bigints.nim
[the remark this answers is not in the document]
dlesnoff: Fixed with new commit.

dlesnoff:
I have changed the modular exponentiation so that it uses an if..elif..else construct rather than only ifs.
I optimized a bit by using the isZero function rather than a modulus == 0.
I added some tests for two other edge cases (0^0 mod n == 1 and 0^1 mod n == 0).
I have tried to rebase on latest commit but I did this uncorrectly (I merged and rebased, while I should have done a simple rebase).

dlesnoff:
Fix PR's history. The 5 different commits might be too much, wich ones should I keep ?

konsumlamm:
> Fix PR's history. The 5 different commits might be too much, wich ones should I keep ?

Don't worry, PRs get squashed anyway, it doesn't really matter how many commits your PR has.

src/bigints.nim
```diff
@@ -956,3 +956,54 @@ iterator `..<`*(a, b: BigInt): BigInt =
...
+    q = r0 div r1
+    rt = r0
+    ut = u
+    r0 = r1
+    u = u1
+    r1 = rt - q * r1
+    u1 = ut - q * u1
+  if r0 != one:
+    raise newException(ValueError, $a & " has no modular inverse")
+  result = ((u mod modulus) + modulus) mod modulus
```
konsumlamm, line 982:
```suggestion
  var
    r0 = a
    r1 = modulus
    s0 = one
    s1 = zero
  while r1 > 0:
    let
      q = r0 div r1
      rk = r0 - q * r1
      sk = s0 - q * s1
    r0 = r1
    r1 = rk
    s0 = s1
    s1 = sk
  if r0 != one:
    raise newException(ValueError, $a & " has no modular inverse")
  result = ((s0 mod modulus) + modulus) mod modulus
```
I renamed u to s, since that's the name used pretty much everywhere. We should add tests for negative `a`.

src/bigints.nim
```diff
@@ -956,3 +956,54 @@ iterator `..<`*(a, b: BigInt): BigInt =
...
+  ## The return value is always in the range `[0, modulus-1]`.
+  if modulus < 0:
+    raise newException(ValueError, "modulus must be strictly positive")
+  elif modulus.isZero:
+    raise newException(DivByZeroDefect, "modulus must be nonzero")
+  elif modulus == 1:
+    return zero
+  elif exponent < 0:
+    if base.isZero:
+      return zero
```
konsumlamm, line 995: I think there's nothing speaking against moving this check up a level, since we have it in the code anyway. Besides, is this really the common behavior? For normal integers, `0 ^ exponent` is undefined if `exponent` is negative.

src/bigints.nim
```diff
@@ -956,3 +956,54 @@ iterator `..<`*(a, b: BigInt): BigInt =
...
+    else:
+      let baseInv = invmod(base, modulus)
+      return powmod(baseInv, -exponent, modulus)
+  else:
+    var
+      exp = exponent
+      basePow = ((base mod modulus) + modulus) mod modulus # Base stays in [0, m-1]
+    result = one
+    while exp != zero:
+      if (exp and one) != zero:
```
konsumlamm, line 1005:
```suggestion
    while not exp.isZero:
      if (exp.limbs[0] and 1) != 0:
```
since `exp` is not zero, `exp.limbs.len` is at least one. We can just do a `uint32` `and` instead of a costly `BigInt` `and`.

konsumlamm (commented):
Now I'm done with reviewing. It took a little longer than expected.

src/bigints.nim
dlesnoff: Negative exponent for modular exponentiation is really common. `0 ^ exponent = 0` for all `exponent != 1` is a general convention. It is common to make the convention `0^{-1} = 0`, it enables to consider functions like `x -> x ^ p-1 mod p` as bijections for all elements of Z/pZ (0 included) in mathematics. But your concern is justified. `0 ^ -n = (0 ^ -1) ^ n` would be the inverse of `0` (raised to exponent `n`) for which I have raised an exception in invmod.
 It makes the code inconsistent.
Moving this check up a level means we do an additional check for the most common case (base > 0, exponent > 0), whereas we optimize an exceptional case (base = 0) that should almost never happen. base.isZero check is not totally free though (it uses a whole loop as of yet).

konsumlamm: I see your point, but note that the loop in `isZero` should run at most once (if the number is correctly normalized). Feel free to mark this as resolved.

src/bigints.nim
dlesnoff: We should also add tests to check that an exception is correctly raised when `a` has no modular inverse.
For negative `a`, we might want to bring it back into the range [0, modulus-1] with :
```nim
var r0 = ((a mod modulus) + modulus) mod modulus
```

src/bigints.nim
dlesnoff: I will finally raise an exception for negative exponent of 0, as 0 is not invertible. I wonder if I should raise an exception inside powmod, or let invmod raise the exception (this would avoid the check).

konsumlamm: I think letting `invmod` raise the exception is fine.

dlesnoff:
I believe there are some cases where invmod returns negative integers depending on the sign of the input. This needs more testing.

src/bigints.nim
```diff
@@ -956,3 +956,64 @@ iterator `..<`*(a, b: BigInt): BigInt =
...
+  ## Compute the modular inverse of `a` modulo `modulus`. 
+  ## The return value is always in the range `[1, modulus-1]`
+  runnableExamples:
+    invmod(3.initBigInt, 7.initBigInt) = 5.initBigInt
+
+  # extended Euclidean algorithm
+  if modulus < 0:
+    raise newException(ValueError, "modulus must be strictly positive")
+  elif modulus.isZero:
+    raise newException(DivByZeroDefect, "modulus must be nonzero")
```
konsumlamm, line 970:
```suggestion
  if modulus.isZero:
    raise newException(DivByZeroDefect, "modulus must be nonzero")
  elif modulus < 0:
    raise newException(ValueError, "modulus must be strictly positive")
```
This way we avoid calling a comparison function.

src/bigints.nim
```diff
@@ -956,3 +956,64 @@ iterator `..<`*(a, b: BigInt): BigInt =
...
+    result = ((s0 mod modulus) + modulus) mod modulus
+
+func powmod*(base, exponent, modulus: BigInt): BigInt =
+  ## Compute modular exponentation of `base` with power `exponent` modulo `modulus`.
+  ## The return value is always in the range `[0, modulus-1]`.
+  runnableExamples:
+    let two = 2.initBigInt
+    let three = 3.initBigInt
+    let seven = 7.initBigInt
+    assert powmod(two, three, seven) == one
```
konsumlamm, line 999:
```suggestion
  runnableExamples:
    assert powmod(2.initBigInt, 3.initBigInt, 7.initBigInt) == 1.initBigInt
```
I think it's more readable this way (besides, `one` is not in scope here I think).

src/bigints.nim
```diff
@@ -956,3 +956,64 @@ iterator `..<`*(a, b: BigInt): BigInt =
...
+  if modulus < 0:
+    raise newException(ValueError, "modulus must be strictly positive")
+  elif modulus.isZero:
+    raise newException(DivByZeroDefect, "modulus must be nonzero")
+  elif modulus == 1:
+    return zero
+  elif exponent < 0:
+    let baseInv = invmod(base, modulus)
+    return powmod(baseInv, -exponent, modulus)
```
konsumlamm, line 1008: We should just reassign `base` and `exponent` here, to avoid repeating the `modulus` checks.

src/bigints.nim
```diff
@@ -956,3 +956,64 @@ iterator `..<`*(a, b: BigInt): BigInt =
...
+  ## The return value is always in the range `[0, modulus-1]`.
+  runnableExamples:
+    let two = 2.initBigInt
+    let three = 3.initBigInt
+    let seven = 7.initBigInt
+    assert powmod(two, three, seven) == one
+  if modulus < 0:
+    raise newException(ValueError, "modulus must be strictly positive")
+  elif modulus.isZero:
+    raise newException(DivByZeroDefect, "modulus must be nonzero")
```
konsumlamm, line 1003:
```suggestion
  if modulus.isZero:
    raise newException(DivByZeroDefect, "modulus must be nonzero")
  elif modulus.isNegative:
    raise newException(ValueError, "modulus must be strictly positive")
```
Same as above, avoids calling a comparison function.

src/bigints.nim
```diff
@@ -956,3 +956,64 @@ iterator `..<`*(a, b: BigInt): BigInt =
...
+  ## The return value is always in the range `[1, modulus-1]`
+  runnableExamples:
+    invmod(3.initBigInt, 7.initBigInt) = 5.initBigInt
+
+  # extended Euclidean algorithm
+  if modulus < 0:
+    raise newException(ValueError, "modulus must be strictly positive")
+  elif modulus.isZero:
+    raise newException(DivByZeroDefect, "modulus must be nonzero")
+  elif a == 0:
```
konsumlamm, line 971:
```suggestion
  elif a.isZero:
```

src/bigints.nim
dlesnoff: I have fixed this, but my solution uses ternary operators. I fear this will not please some people here … I will fix this but I need to sleep and eat.

src/bigints.nim
```diff
@@ -956,3 +956,63 @@ iterator `..<`*(a, b: BigInt): BigInt =
...
+    raise newException(DivByZeroDefect, "modulus must be nonzero")
+  elif modulus.isNegative:
+    raise newException(ValueError, "modulus must be strictly positive")
+  elif modulus == 1:
+    return zero
+  else:
+    var
+      base = base
+      exponent = exponent
+    if exponent.isNegative == true:
```
konsumlamm, line 1007: This should stay `exponent < 0`, since zero can also be negative (or change to `not exponent.isZero and exponent.isNegative`). The `== true` is unnecessary either way.

konsumlamm (approved):

dlesnoff:
Tried to rebase, then merge another PR. History is a mess. Please squash my commits before merging.

dlesnoff:
I removed an example that used a custom modular exponentiation similar to my powmod function. I could have replaced with it with just one call to the function, it would not have been interesting.

konsumlamm:
@dlesnoff can you rebase this, to resolve the conflicts?
@narimiran can we merge this soon, to avoid more conflicts? Unless you have any objections of course.

dlesnoff:
merged the changes to resolve the conflicts.

konsumlamm (approved):

PR #66, opened 2022-01-07
Add gcd and abs with tests

dlesnoff:
Implemented gcd with least absolute remainder (LAR) method.
I also implemented classical naïve gcd, and compared the two. Gcd tests are about twice faster with LAR so I kept only LAR method.
Implemented abs as max(a,-a).

### DIFF
--- a/src/bigints.nim
+++ b/src/bigints.nim
@@ -960,7 +960,9 @@ iterator `..<`*(a, b: BigInt): BigInt =
     yield res
     inc res
 
-func abs*(a: BigInt): BigInt = max(a, -a)
+func abs*(a: BigInt): BigInt =
+  result = a
+  result.isNegative = false
 
 func gcd*(a, b: BigInt): BigInt =
   ## Returns the greatest common divisor of `a` and `b`.
--- a/src/bigints.nim
+++ b/src/bigints.nim
@@ -959,3 +959,37 @@ iterator `..<`*(a, b: BigInt): BigInt =
   while res < b:
     yield res
     inc res
+
+func abs*(a: BigInt): BigInt = max(a, -a)
+
+func gcd*(a, b: BigInt): BigInt =
+  ## Returns the greatest common divisor of `a` and `b`.
+  if a.isZero:
+    return b
+  elif b.isZero:
+    return a
+  elif a < 0:
+    if b < 0:
+      return gcd(-a, -b)
+    else:
+      return gcd(-a, b)
+  elif b < 0:
+    return gcd(a, -b)
+  else:
+    var
+      u: BigInt = a
+      v: BigInt = b
+      t: BigInt = b
+      r: BigInt = a mod b
+    while v != 0:
+      t = v
+      r = u mod v
+      # v = min(r, r-v)
+      if r > (v shr 1):
+        v = v - r
+      else:
+        v = r
+      u = t
+    # We return abs(u)
+    result = abs(u)
+    
--- a/tests/tbigints.nim
+++ b/tests/tbigints.nim
@@ -379,5 +379,26 @@ template main() =
     doAssert pow(zero, 0) == one
     doAssert pow(zero, 1) == zero
 
+  block: # gcd
+    let a = "866506".initBigInt
+    let b = "140640".initBigInt
+    let two = 2.initBigInt
+    doAssert gcd(a, b) == two
+    # gcd(a, -b) = gcd(a, |b|)
+    doAssert gcd(-a, b) == two
+    doAssert gcd(a, -b) == two
+    doAssert gcd(-a, -b) == two
+
+  block: # properties of gcd
+    let a = "668403".initBigInt
+    let b = "753160".initBigInt
+    let c = "249115".initBigInt
+    doAssert gcd(a, b) == gcd(b, a)
+    doAssert gcd(a, zero) == a
+    doAssert gcd(a, a) == a
+    doAssert gcd(c*a, c*b) == c*gcd(a,b)
+    doAssert gcd(a, gcd(b, c)) == gcd(gcd(a, b), c)
+    doAssert gcd(a, b) == gcd(b, a mod b)
+
 static: main()
 main()


src/bigints.nim
```diff
@@ -959,3 +959,35 @@ iterator `..<`*(a, b: BigInt): BigInt =
   while res < b:
     yield res
     inc res
+
+func abs*(a: BigInt): BigInt = max(a,-a)
+
+func gcd*(a, b: BigInt): BigInt =
+  ## Returns the greatest common divisor of `a` and `b`
```
xflywind, line 966:
```suggestion
  ## Returns the greatest common divisor of `a` and `b`.
```

src/bigints.nim
```diff
@@ -959,3 +959,35 @@ iterator `..<`*(a, b: BigInt): BigInt =
...
+  if a == 0:
+    return b
+  if b == 0:
+    return a
+  if a < 0:
+    if b < 0:
+      return gcd(-a,-b)
```
dlesnoff, line 973: Reminder : add space after comma.

src/bigints.nim
```diff
@@ -959,3 +959,35 @@ iterator `..<`*(a, b: BigInt): BigInt =
   while res < b:
     yield res
     inc res
+
+func abs*(a: BigInt): BigInt = max(a,-a)
```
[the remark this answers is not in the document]
dlesnoff: 963 and 973 spaces have been added in latest push.

src/bigints.nim
dlesnoff: Fixed this punctuation issue. Thanks.

src/bigints.nim
```diff
@@ -959,3 +959,37 @@ iterator `..<`*(a, b: BigInt): BigInt =
   while res < b:
     yield res
     inc res
+
+func abs*(a: BigInt): BigInt = max(a, -a)
```
konsumlamm, line 963:
```suggestion
func abs*(a: BigInt): BigInt =
  result = a
  result.isNegative = false
```
`BigInt`s are already stored as sign and absolute value.

src/bigints.nim
```diff
@@ -959,3 +959,37 @@ iterator `..<`*(a, b: BigInt): BigInt =
...
+    return b
+  elif b.isZero:
+    return a
+  elif a < 0:
+    if b < 0:
+      return gcd(-a, -b)
+    else:
+      return gcd(-a, b)
+  elif b < 0:
+    return gcd(a, -b)
```
konsumlamm, line 977: I'd rewrite this to make local copies of `abs(a)` and `abs(b)`, rather than calling the function again.

src/bigints.nim
```diff
@@ -959,3 +959,37 @@ iterator `..<`*(a, b: BigInt): BigInt =
...
+  else:
+    var
+      u: BigInt = a
+      v: BigInt = b
+      t: BigInt = b
+      r: BigInt = a mod b
```
konsumlamm, line 983: The type annotations are unnecessary here, type inference is good enough for that.

src/bigints.nim
```diff
@@ -959,3 +959,37 @@ iterator `..<`*(a, b: BigInt): BigInt =
...
+    while v != 0:
+      t = v
+      r = u mod v
+      # v = min(r, r-v)
+      if r > (v shr 1):
+        v = v - r
+      else:
+        v = r
+      u = t
```
konsumlamm, line 992:
```suggestion
    var
      u: BigInt = a
      v: BigInt = b
    while v != 0:
      let
        t = v
        r = u mod v
      # v = min(r, r-v)
      if r > (v shr 1):
        v = v - r
      else:
        v = r
      u = t
```

konsumlamm:
> Implemented gcd with least absolute remainder (LAR) method.

Do you have a link to that algorithm? I didn't really find anything on it.

We should also consider using [binary GCD](https://en.wikipedia.org/wiki/Binary_GCD_algorithm) (it doesn't use division/modulo, which I imagine is quite expensive for `BigInt`s).

dlesnoff:
> Do you have a link to that algorithm? I didn't really find anything on it.

https://en.wikipedia.org/wiki/Euclidean_algorithm#Implementations (there is a paragraph on it, see Method of least absolute remainder).
I should not have given it an abbreviation, it is a small modification of the Euclid’s algorithm, so it is more an optimization than a full algorithm itself. This method uses as few steps as possible for the euclidean algorithm and should take fewer steps than Stein’s algorithm (binary GCD). But indeed, the operations used should be compared between the variants.
The multiplication, division and modulo operation should be improved first, before we do any optimizations or benchmarks on gcd.

src/bigints.nim
dlesnoff: I try to avoided initialisations inside the while loop. But we indeed have immutable variables with this let construct. Are you sure this version is better ? Do we not make more copies that way ?

src/bigints.nim
dlesnoff: I forgot to remove typehints after copy-paste from C code …

src/bigints.nim
konsumlamm: I'm pretty sure we don't make more copies, assignment copies anyway.

src/bigints.nim
```diff
@@ -959,3 +959,39 @@ iterator `..<`*(a, b: BigInt): BigInt =
...
+      t: BigInt = b
+      r: BigInt = a mod b
+    while v != 0:
+      t = v
+      r = u mod v
+      # v = min(r, r-v)
+      if r > (v shr 1):
+        v = v - r
+      else:
+        v = r
```
konsumlamm, line 993: If i read the description of the least absolute remainders method correctly, this is not what the method describes. It says that the condition you check for is never satisfied.

konsumlamm:
> This method uses as few steps as possible for the euclidean algorithm and should take fewer steps than Stein’s algorithm (binary GCD). But indeed, the operations used should be compared between the variants. The multiplication, division and modulo operation should be improved first, before we do any optimizations or benchmarks on gcd.

You already did optimizations (using the method of least absolute remainders) and benchmarks, so I don't understand your point.

While the euclidean algorithm (and the method of least absolute remainders) uses less steps than binary GCD, the latter does not use division, which is slow for `BigInt`s. While we can (and should) optimize division, I doubt that'll change that division is comparatively slow.

src/bigints.nim
dlesnoff: I wonder if I can make a commit through a comment.
```nim
var
  a = abs(a)
  b = abs(b)
```

src/bigints.nim
dlesnoff: if the remainder is greater than half of v, then we need to take the negative remainder. It happens to be equal to `v - r`.

dlesnoff: The `v` we get is always smaller than `v/2`, hence it never satisfies the condition I check. I think the confusion comes from the different values `v` has before or after assignation. Wikipedia probably gives the mathematical result of the algorithm.
The tests are correct and find same results as gcd. My tests are not great, I believe to have only tested on single-precision numbers.

src/bigints.nim
dlesnoff: We do make more initializations. I will try to test both versions.

konsumlamm:
> Implemented gcd with least absolute remainder (LAR) method. I also implemented classical naïve gcd, and compared the two. Gcd tests are about twice faster with LAR so I kept only LAR method.

@dlesnoff could you share your benchmarks? I did some (very) naive benchmarks myself to compare the standard euclidean algorithm, the method of least absolute remainders and binary GCD. My results say that binary GCD is *a lot* faster (about 5x), however, I didn't see the method of least absolute remainders being twice as fast as the euclidean algorithm.

dlesnoff:
I have find Knuth’s seminumerical algorithms (The Art of Computer Programming vol.2) in the library, and here is its opinion on the subject :
> If `u` and `v` are very large integers, requiring a multiple-precision representation, the binary method (Algorithm B) is a simple and fairly efficient means of calculating their greatest common divisor, since it involves only subtractions and shifting.
> By contrast, Euclid’s algorithm seems much less attractive, since step A2 requires a multiple-precision division of `u` by `v`. But this difficulty is not really as bad as it seems, since we will prove in Section 4.5.3 that the quotient `⌊u/v⌋ ` is almost always very small. […] assuming random inputs […] the quotient will be less than 1000 approximately 99.856% of the time. Furthermore […] we don’t really mind having a large quotient `q` since Euclid’s algorithm makes a great deal of progress when it replaces `u` by `u mod v` in such a case.
It then mentions an amelioration by D.H.Lehmer but binary gcd algorithm can be modified in a way similar to Lehmer’s (we do virtual computation) that make binary gcd faster.
So I guess you are right.

dlesnoff:
> > Implemented gcd with least absolute remainder (LAR) method. I also implemented classical naïve gcd, and compared the two. Gcd tests are about twice faster with LAR so I kept only LAR method.
> 
> @dlesnoff could you share your benchmarks?
I simply ran the tests for gcd (classical euclid algorithm, same code without the if-else) and gcdx (the one in the PR, that I called gcdx before).
As you can see, the integers in the tests are not multi-precision. So I have not really "benchmarked" it. I might rewrite classical gcd of course if you want, but feel free to make a PR with your binary gcd and I will close this one.

PR #71, opened 2022-01-09
Add fastLog2 operation, closes #67

### LINKED ISSUE #67: Add significant bits/fastLog2 operation for BigInts
In order to improve modular exponentiation or to write other higher-order functions (functions that do not rely on BigInts internal representation), we might want to count the number of significant bits (bits after leading zeros, without the sign bit in complement representation of signed integers) or at least the logarithm in base 2 of the number.

We can determine the number of significant bits from the logarithm in base 2 as:
significantBits(n) = floor(log_2(n)) + 1
```Nim
func significantBits*(n: BigInt): int =
  if n < 0:
    return significantBits(-n)
  fastLog2(n.limbs.high) + 32*(n.limbs.len-1)
```
How would you optimize the negative case ? Can we just take the two’s complement of the highest limb and then compute the fastLog2(n.limbs.high) ?
```Nim
func significantBits*(n: BigInt): int =
  var highest_limb = n.limbs.high
  if n < 0:
    highest_limb = not n.limbs.high + 1
  fastLog2(highest_limb) + 32*(n.limbs.len-1)
```

dlesnoff:
Add a bitops-like operation for BigInts with respective test.
Solves #67.


src/bigints.nim
```diff
@@ -959,3 +959,5 @@ iterator `..<`*(a, b: BigInt): BigInt =
   while res < b:
     yield res
     inc res
+
+func fastLog2*(a: BigInt): int = bitops.fastLog2(a.limbs[a.limbs.high]) + 32*(a.limbs.len-1)
```
narimiran, line 963: Minor: Put the body of the proc in a separate line, and add doc comment.

narimiran:
Ok, maybe the implementation is fast, but should that be reflected in the name of the function? Why not just `log2`?

dlesnoff:
I do not want to say that my implementation is fast, but I want the user to understand this is the equivalent for BigInt of the `fastLog2` function of the `std/bitops` library (that usually operates on SomeInteger).

konsumlamm:
Hmm, I wonder what would be best for negative numbers. I think calculating log2 of the absolute value makes sense, but then it should at least be documented. Alternatively, we could throw an exception, but I don't know how useful that would be.

src/bigints.nim
```diff
@@ -1068,6 +1068,13 @@ iterator `..<`*(a, b: BigInt): BigInt =
     yield res
     inc res
 
+func fastLog2*(a: BigInt): int =
+  ## Computes the logarithm in base 2 of `a`.
+  ## If `a` is negative, returns the logarithm of `abs(a)`.
```
konsumlamm, line 1073:
```suggestion
  ## If `a` is negative, returns the logarithm of `abs(a)`.
  ## If `a` is zero, returns -1.
```

src/bigints.nim
```diff
@@ -1068,6 +1068,14 @@ iterator `..<`*(a, b: BigInt): BigInt =
...
+  ## If `a` is zero, returns -1.
+  if a.isZero:
+    return -1
+  bitops.fastLog2(a.limbs[a.limbs.high]) + 32*(a.limbs.len-1)
```
narimiran, line 1077: `a.limbs.high` and `a.limbs.len-1` are the same thing. Why use both, and not just one of those variants?

tests/tbigints.nim
```diff
@@ -385,6 +385,20 @@ proc main() =
     doAssert "fedcba9876543210".initBigInt(base = 16) == b
     doAssert "ftn5qj1r58cgg".initBigInt(base = 32) == b
 
+  block: # fastLog2
+    let a = one shl 31
```
narimiran, line 389: We need some "not-power-of-two" tests too.

dlesnoff: Do you have any idea of easily checkable tests then ?
I have tested the fastLog of a+b which is not a power of 2.
I have added one test but it is not obvious what is it’s logarithm.
There are not much possible errors with this simple function and I am tired of waiting 30 seconds every time I have to test my code. It is much longer now with all the GC’s.
Maybe some factorial, since with Stirling approximation, we can guess the number of digits ahead ?

konsumlamm: I'd also test values with all 1's, e.g. `one shl 64 n- 1` and a few negative non-powers-of-two.

As for running the tests yourself, you can just `nim r tests/tbigints.nim`. (In theory, the behavior should be the same for all GCs, but in practice, there are bugs, as exemplified by #88).

narimiran: > Do you have any idea of easily checkable tests then ?

I would do some sanity checks, e.g. log of 7, 8, 9, 24, 32, 48, etc., just to make sure it works for the basics. And then do what @konsumlamm suggests.

dlesnoff:
Are the new tests correct, enough, sufficient ?

konsumlamm:
> Are the new tests sufficient ?

Have you read https://github.com/nim-lang/bigints/pull/71#discussion_r794466004 and https://github.com/nim-lang/bigints/pull/71#discussion_r794624698?

dlesnoff:
I have, I added tests corresponding to these comments and I am asking if the new tests are okay for you.

konsumlamm:
Huh, for some reason your latest commit doesn't shop up here... Maybe you can try to rebase/merge the master branch.

dlesnoff:
I was 10 commits behind master. That might be the cause.
Commit did show up in the history though (but not in diff), so I was quite surprised by your answer.
The diff is now correct.

konsumlamm (approved):

Issue #72, opened 2022-01-09
Type conversion for small BigInts into integers


dlesnoff:
Since BigInt's limb parameter has become private, thus we can not easily cast BigInts with absolute value less than `2^32` to an `int` or even a `uint32`. 
The only workaround yet is to use the keyword `cast[int]`. But this does not always do what we want.
To get only the smallest limb, we also have to define a mask like `2^32 - 1` and to apply an ineffective `and`.
Can we implement some functions like … ?:
- `toUint32(a: BigInt): uint32` returns the only limb for BigInts between 1-2^32 and +2^32-1 (small BigInts).
-  `toInt(a: BigInt): int` cast the BigInt if it is small enough, into an int, according to the sign of BigInt, with correct representation (2's complement) and raises a ValueError if the cast can not be done (if int is stored on 32 bits, it may not fit if the heavy weight bit of the uint32 corresponding to first limb is 1).
- `toUint64(a: BigInt): uint64` returns the uint64 made up of one to two limbs, if the BigInt is small enough.
-  `toFloat` like described in the similar issue #34.

konsumlamm:
> The only workaround yet is to use the keyword `cast[int]`. But this does not always do what we want.

That never does what you want. I suggest you read the [documentation for `cast`](https://nim-lang.org/docs/manual.html#statements-and-expressions-type-casts), it merely reinterprets the underlying bits.

I generally agree that functions to convert a `BigInt` to some integer types are useful, but how should we handle the case of the `BigInt` being out of range? I see several possibilities:
* raise an exception
* return an [`Option`](https://nim-lang.org/docs/options.html#Option)
* return the lowest/highest value

narimiran:
@Araq: you might be interested in this discussion, i.e. give your opinion on how to best proceed with this.

dlesnoff:
We can not return the lowest/highest value for type conversions to integers as a BigInt can be any value that an int can contain.
The first solution forces the user to wrap any potentially dangerous cast in a `try .. except .. finally` clause.
I have not used `Option` in my programs yet, if I understand the user will have to check (everytime ?) whether the Option contains a value ?

konsumlamm:
This issue should be reopened, `toSignedInt` only provides conversion to signed integers. I'd also argue that it should be renamed to `toInt` and also support unsigned integers (rather than adding a new `toUnsignedInt` function).

Araq:
I also really need `pred`  and `succ`. Implementation can be as simple as:

```nim

[...]

PR #73, opened 2022-01-09
Examples: Fix compiling issues and imports, closes #69

### LINKED ISSUE #69: Some examples do not compile
I have tested the examples but could not compile some of them probably due to some optional dependencies like unsigned in the pidigits example.
Is it an old file of the library that has been removed, or an external library ?
In the latter case, it should be added to nimble optional dependencies list if nimble has such an option.

Furthermore, some syntax has become obsolete like `.. <` that is now `..<`.

dlesnoff:
Fix compilation errors for many examples.
Change `import bigints` to `import ../src/bigints` to ensure examples are tested with the development version and not the stable version. This might be changed back when a stable version is going to be released.
Changed imports from standard library to differentiate them from external libraries :
`import strutils, math` become `import std/[strutils, math]`
As there is no implicit type conversion for integers to BigInt anymore, operations between a BigInt and an Integer now become operations between BigInt and integer explicitly converted to BigInt.
Now three examples do not work :

- [x] pidigits.nim - rangeDefect
- [x] rc_pidigits.nim - rangeDefect (almost identical to pidigits.nim)
- [x] rc_godtheinteger.nim - infinite Loop (takes more than 30 seconds)
- [x] rc_leftfactorials.nim - depends on an external library iterutils.
- [x] rc_paraffins.nim


pietroppeter:
https://github.com/nim-lang/bigints/issues/69#issuecomment-1008427337

dlesnoff:
Sorry I have seen your comment after my PR, I am going to rebase my fix.

pietroppeter:
>  rc_godtheinteger.nim - infinite Loop (takes more than 30 seconds)

see http://rosettacode.org/wiki/9_billion_names_of_God_the_integer#Nim that version is supposed to be slow and as such should not be kept running. version 2 is the fast one.

dlesnoff:
I have :
- changed imports back to `import bigint`
- added a `nim.cfg`
- in src/bigints.nim, there is now a `^` decorator for `pow`. Why was this removed ?
It helps to keep some examples easy to read in my opinion.
- I can not compile anymore rc_parafins.nim. It fails at line 42. I am unsure yet how to fix it.
- rc_pidigits.nim and pidigits.nim are still broken, it would take me a lot of time to understand the algorithms and rewrite them.
- @def- Is it possible to replace the slice iterator from iterutils package in your example, to remove iterutils dependency ?

narimiran:
@konsumlamm, your review please :)

pietroppeter:
>  Is it possible to replace the slice iterator from iterutils package in your example, to remove iterutils dependency ?

I would adapt it to something like this ([playground](https://play.nim-lang.org/#ix=3LMp)):

```nim
iterator square: int =
  var i = 0
  while true:
    yield i*i
    inc i

var i = 0
for n in square():
  if i <= 10:
    echo i, ": ", n
  elif i <= 110 and i mod 10 == 0:
    echo i, ": ", n
  elif i >= 1000 and i <= 10_000 and i mod 1000 == 0:
    echo i, ": ", n
  elif i > 10_000:
    break
  inc i
```

pietroppeter:
in `examples/rc_godtheinteger.nim` you could change this line:

```nim
for x in [23, 123, 1234, 12345]:
```
to

```nim
# for 12345 this implementation is too slow, for a faster implementation see rc_godtheinteger2.nim
for x in [23, 123, 1234]:
```

dlesnoff:
Thanks a lot @pietroppeter, the two new commits follow up from your ideas.

examples/pidigits.nim
```diff
@@ -20,16 +29,16 @@ proc extractDigit(): int32 =
   if tmp2 >= den:
     return -1
 
-  result = int32(tmp1.limbs[0])
```
dlesnoff, line 23: This example can not be rewritten as long as we have not defined a way to convert a BigInt into an int, see issue #72. Type cast does not do the same thing, it interprets the BigInt as an int ! So the first limb is understood as the 2’s complement representation of another number.

src/bigints.nim
```diff
@@ -455,6 +455,8 @@ func pow*(x: BigInt, y: Natural): BigInt =
     exp = exp shr 1
     base *= base
 
+func `^`*(x: BigInt, y: Natural): BigInt = pow(x,y)
```
konsumlamm, line 458:
```suggestion
func `^`*(x: BigInt, y: Natural): BigInt = pow(x, y)
```
While I have nothing against this addition, I don't think that a PR that's (supposedly) about fixing examples should add new features. I'd prefer a separate Pr for this.

examples/elliptic.nim
```diff
@@ -1,19 +1,25 @@
...
+  zero = 0.initBigInt
 
 proc `^`(base: int; exp: int): BigInt =
   let base = base.initBigInt
   var exp = exp
-  result = 1.initBigInt
+  result = one
   while exp > 0:
     result *= base
     dec(exp)
```
konsumlamm, line 17: I'd rewrite this whole function to just use `bigints.pow`

examples/elliptic.nim
```diff
@@ -1,19 +1,25 @@
...
 
 let
-  Pcurve: BigInt = 2^256 - 2^32 - 2^9 - 2^8 - 2^7 - 2^6 - 2^4 - 1
+  Pcurve: BigInt = 2^256 - 2^32 - 2^9 - 2^8 - 2^7 - 2^6 - 2^4 - one
   N = initBigInt("FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEBAAEDCE6AF48A03BBFD25E8CD0364141", 16)
-  Acurve = 0.initBigInt
+  Acurve = zero
   Bcurve = 7.initBigInt
   Gx = initBigInt("55066263022277343669578718895168534326250603453777594175500187360389116729240")
   Gy = initBigInt("32670510020758816978083085130507043184471273380659243275938904335757337482424")
```
konsumlamm, line 25: We should really change the names of those variables

examples/pidigits.nim
```diff
@@ -1,10 +1,19 @@
...
 
-import os, strutils, unsigned, bigints
+import std/[os, strutils]
+import bigints
+
+const
+  one = 1.initBigInt
+  zero = 0.initBigInt
+  ten = 10.initBigInt
+  two = 2.initBigInt
```
konsumlamm, line 10:
```suggestion
  zero = 0.initBigInt
  one = 1.initBigInt
  two = 2.initBigInt
  ten = 10.initBigInt
```
Let's at least order these by *something*.

examples/rc_godtheinteger2.nim
```diff
@@ -2,9 +2,11 @@
 import bigints
 
 var p = @[1.initBigInt]
+const
+  zero = 0.initBigInt
```
konsumlamm, line 6: Defining a `const` isn't really worth it when it's only used once.

examples/rc_modexp.nim
```diff
@@ -2,19 +2,26 @@
...
+    if e mod two == one:
       result = (result * b) mod m
-    e = e div 2
+    e = e div two
     b = (b.pow 2) mod m
 
 var
   a = initBigInt("2988348162058574136915891421498819466320163312926952423791023078876139")
   b = initBigInt("2351399303373464486466122544523690094744975233415544072992656881240319")
+  res = powmod(a, b, initBigInt(10).pow 40)
```
konsumlamm, line 24:
```suggestion
  res = powmod(a, b, initBigInt(10).pow(40))
```

konsumlamm (commented):
In the future, we should also test these examples in the CI (I didn't even know they existed until recently).

src/bigints.nim
dlesnoff: I will remove this. It might break some cosmetic changes I made in the examples elliptic.nim and rc_pow.nim.

dlesnoff:
All examples compile, only pidigits examples does not work but this may be fixed with PR #74.

dlesnoff:
Pidigits examples now compile, but in case toSignedInt change name, we will have to modify slightly the examples again.
In pidigits.nim, I added command line basic parsing. I thought it was broken, because I did not provide an argument, and nothing explained that an argument was expected.

dlesnoff:
@konsumlamm or @pietroppeter may I ask for a review on this PR ?

examples/elliptic.nim
```diff
@@ -35,25 +35,25 @@ proc modinv(a: BigInt): BigInt =
     highm = temp
     swap hm, lm
     swap highm, lowm
-  result = lm mod Pcurve
+  result = lm mod primeCurve
```
konsumlamm, line 38: This can be replaced by `invmod(a, primeCurve)`.

examples/rc_combperm.nim
```diff
@@ -1,19 +1,22 @@
 # Solution for http://rosettacode.org/wiki/Combinations_and_permutations
 import bigints
 
+const
+  zero = 0.initBigInt
```
konsumlamm, line 5: This `const` is used once, just inline the `0.initBigInt`.

examples/pidigits.nim
```diff
@@ -1,10 +1,19 @@
 # Translation of http://benchmarksgame.alioth.debian.org/u32/program.php?test=pidigits&lang=go&id=4
```
konsumlamm, line 1: This link is broken.

examples/rc_modexp.nim
```diff
@@ -2,19 +2,26 @@
...
-  result = initBigInt(1)
-  while e > 0:
-    if e mod 2 == 1:
+  result = one
+  while e > zero:
+    if e mod two == one:
       result = (result * b) mod m
-    e = e div 2
+    e = e div two
     b = (b.pow 2) mod m
```
konsumlamm, line 19: This can just use our `powmod` now.

konsumlamm (commented):
Can you change the link in `examples/rc_godtheinteger.nim` to `https://rosettacode.org/wiki/9_billion_names_of_God_the_integer` and change all links to https?

examples/pidigits.nim
konsumlamm: Also, isn't this example essentially a duplicate of `rc_pidigits.nim`?

dlesnoff: The pidigits.nim file is the interactive version of rc_pidigits.nim
The pidigits.nim asks the user for the number of digits of pi he wants, whereas rc_pidigits.nim prints indefinitely new digits of pi.
I can remove rc_pidigits.nim if you want. Especially knowing that the code (now outdated) is published on the website.

examples/rc_modexp.nim
dlesnoff: I have deleted this example in powmod PR. I will remove it in this PR too.

examples/pidigits.nim
konsumlamm: I don't really care which one we keep.

konsumlamm (approved):

dlesnoff:
@narimiran May I ask to merge this PR ?

PR #74, opened 2022-01-12
add `toSignedInt`, closes #72

### LINKED ISSUE #72: Type conversion for small BigInts into integers

narimiran:


src/bigints.nim
```diff
@@ -753,6 +753,45 @@ func `divmod`*(a, b: BigInt): tuple[q, r: BigInt] =
...
+      result = some(T(x.limbs[0]))
+    if result.isSome and x.isNegative:
+      if result.get == int.low: return # a hack to prevent underflow
+      result = some(not(result.get - 1))
+  else:
+    if x.limbs.len > 1:
+      result = none(T)
+    else:
+      if x.isNegative:
+        if x.limbs[0] - 1 > uint32(T.high):
```
konsumlamm, line 784: What if `x.limbs[0]` is 0? That's not disallowed currently afaik.

konsumlamm:
Nice! I think we should also support unsigned integers, is there any particular reason you didn't include them here? `toInt` would probably be a better name then, unless we want a separate `toUnsignedInt` function (which would be kinda weird since we already have to specify the int type anyway)?

tests/tbigints.nim
```diff
@@ -384,5 +384,42 @@ template main() =
...
     doAssertRaises(DivByZeroDefect): discard divmod(one, zero)
 
+  block: # toSignedInt
+    let
+      a = initBigInt(7)
+      b = initBigInt(-7)
+      c = initBigInt(int32.high)
+      d = initBigInt(int32.low)
+      e = "123456789123456789123456789".initBigInt
+      f = "340282366920938463463374607431768211456".initBigInt
```
dlesnoff, line 394: Could you add a comment to tell how many limbs are necessary for them to be represented ?
I guess log_2(e) = 87 bits so e needs 3 limbs and 128 bits for f(=2^128) that sums to 4 limbs.

src/bigints.nim
```diff
@@ -753,6 +753,45 @@ func `divmod`*(a, b: BigInt): tuple[q, r: BigInt] =
...
+      b = 130.initBigInt
+    assert toSignedInt[int8](a) == some(44'i8)
+    assert toSignedInt[int8](b) == none(int8)
+    assert toSignedInt[int](b) == some(130)
+
+  when T is int64 or T is int:
+    if x.limbs.len > 2:
+      result = none(T)
+    elif x.limbs.len == 2:
+      result = some((T(x.limbs[1]) shl 32) + T(x.limbs[0]))
```
dlesnoff, line 773: What happens if you try to convert a number that is between int64.high (=2^63-1) and 2^64 ?

src/bigints.nim
```diff
@@ -753,6 +753,45 @@ func `divmod`*(a, b: BigInt): tuple[q, r: BigInt] =
...
+  runnableExamples:
+    import options
+    let
+      a = 44.initBigInt
+      b = 130.initBigInt
+    assert toSignedInt[int8](a) == some(44'i8)
+    assert toSignedInt[int8](b) == none(int8)
+    assert toSignedInt[int](b) == some(130)
+
+  when T is int64 or T is int:
```
dlesnoff, line 769: As I understand the type, `int` uses 32 bit on a 32 bit architecture and 64 bits on a 64 bits architecture (or depending on the compiler’s flags ?). The code under the `when` seems to assume T uses 64 bit and I fear that might not always be the case if `T` is an `int`.

tests/tbigints.nim
```diff
@@ -384,5 +384,42 @@ template main() =
...
+      g = initBigInt(int64.high)
+      h = initBigInt(int64.low)
+      i = initBigInt(int64.low + 1)
```
dlesnoff, line 397: Could we test for int64.high + 1 and int64.low - 1 ?

konsumlamm: That would be over-/underflow, what would you expect the test to do?

src/bigints.nim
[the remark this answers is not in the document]
dlesnoff: Why > 4 ? Can we just not test equality `sizeof(T) == 8` ? (Unless we target 128-bit architecture)

tests/tbigints.nim
dlesnoff: To return a none Option. But as of right now, I believe the code will make an over-/underflow.

src/bigints.nim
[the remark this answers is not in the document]
dlesnoff: As `int` is architecture dependent, the code under the when seems to assume T uses 64 bit and I fear that might not always be the case if T is an int and architecture is 32-bit.

src/bigints.nim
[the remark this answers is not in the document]
dlesnoff: I believe this does not work for integers between int64.high (=2^63-1) and 2^64.

tests/tbigints.nim
konsumlamm: `int64.high + 1` and `int64.low - 1` can't be represented in 64 bits, so I don't know how you expect this to work. I'm currently not sure what the expected behaviour for signed over-/underflow is (wraparound, exception, UB), but either way, it won't be the same as `initBigInt(int64.high) + initBigInt(1)`/`initBigInt(int64.low) - initBigInt(1)`.

src/bigints.nim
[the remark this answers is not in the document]
narimiran: Yeah, testing for `sizeof` is definitely better.

narimiran:
> I think we should also support unsigned integers, is there any particular reason you didn't include them here? 

To get your comments, improve the code, and then based on that, continue with unsigned.

tests/tbigints.nim
narimiran: > Could we test for int64.high + 1 and int64.low - 1 ?

We can.

> what would you expect the test to do?

To return `none`.

src/bigints.nim
narimiran: Can we have a negative zero? *Should* we have a negative zero?

src/bigints.nim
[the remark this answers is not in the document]
narimiran: > e.g. `0xffffffff40000000` gives the wrong result

Now that's a good candidate to include in the tests :)

src/bigints.nim
konsumlamm: Currently yes, for example `-0.initBigInt` (it gets treated the same as any other zero though). Wether that *should* be allowed is a different question.

src/bigints.nim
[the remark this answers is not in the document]
narimiran: See the updated code.

src/bigints.nim
```diff
@@ -753,6 +753,53 @@ func `divmod`*(a, b: BigInt): tuple[q, r: BigInt] =
...
+    assert toSignedInt[int8](b) == none(int8)
+    assert toSignedInt[int](b) == some(130)
+
+  when sizeof(T) == 8:
+    if x.limbs.len > 2:
+      result = none(T)
+    elif x.limbs.len == 2:
+      if x.limbs[1] > uint32.high shr 1:
+        if x.isNegative and x.limbs[0] == 0:
+          result = some(T(int64.low))
```
konsumlamm, line 775: This breaks for e.g. `-initBigInt(0xffffffff00000000'u64)`, you also need to check `x.limbs[1]`.

konsumlamm (commented):
I think it would be easier/cleaner to fully convert the absolute values to unsigned ints (the only checks required for that are checking the length of `x.limbs`) and then convert that to signed ints.

Araq:
I need these additions so I'm merging it. Later on more improvements can follow. I also don't mind to change my code if you decide to tune the API.

PR #75, opened 2022-01-12
Add `gcd` & `abs`

konsumlamm:
Subsumes #66 (see https://github.com/nim-lang/bigints/pull/66#issuecomment-1011242477).

`gcd` is implemented using the [binary GCD algorithm](https://en.wikipedia.org/wiki/Binary_GCD_algorithm).


src/bigints.nim
```diff
@@ -752,6 +760,43 @@ func `divmod`*(a, b: BigInt): tuple[q, r: BigInt] =
...
+    if x == 0:
+      count += 32
+    else:
+      return count + countTrailingZeroBits(x)
+  return count
+
+func gcd*(a, b: BigInt): BigInt =
+  ## Returns the greatest common divisor (GCD) of `a` and `b`.
+  runnableExamples:
+    assert gcd(54.initBigInt, 24.initBigInt) == 6.initBigInt
```
dlesnoff, line 775: I do not know if it's preferable that runnableExamples are fast to execute, but one funny example (to showcase multiple precision integers too) would be :
```nim
let two = 2.initBigInt
assert gcd(two.pow(54) - 1, two.pow(24) - 1) == two.pow(6) - 1
```

konsumlamm: In general it doesn't matter if `runnableExamples` are fast to execute, as they're only run when generating documentation. I'm not entirely convinced that this particular example is suited for the `runnableExamples`, but ofc we can add it to the test suite.

dlesnoff (approved):
Thanks a lot. Might be improved in the future by other methods designed for multiple-precision arithmetic like Lehmer's (Knuth 4.5.2).

src/bigints.nim
dlesnoff: Could you add the test and rebase @konsumlamm ?
Same as with the modular exponentiation, we will also get more conflicts if we let the PR pending.
May you then merge this PR @narimiran ?

PR #79, opened 2022-01-16
Pred succ functions

dlesnoff:
For @Araq, predecessor and successor functions.

### DIFF
--- a/src/bigints.nim
+++ b/src/bigints.nim
@@ -977,6 +977,14 @@ func dec*(a: var BigInt, b: int32 = 1) =
   var c = a
   subtractionInt(a, c, b)
 
+proc pred*(a: BigInt; def: SomeInteger|string = 1): BigInt =
+  ## Decrements `a` by `def`. By default, subtracts one.
+  a - initBigInt(def)
+
+proc succ*(a: BigInt; def: SomeInteger|string = 1): BigInt =
+  ## Increments `a` by `def`. By default, adds one.
+  a + initBigInt(def)
+
 
 iterator countup*(a, b: BigInt, step: int32 = 1): BigInt =
   ## Counts from `a` up to `b` (inclusive) with the given step count.


konsumlamm:
These can be optimized by using `inc`/`dec`, which already take an integer argument, rather than unnecessarily creating a new `BigInt`.

I really don't like having `def: SomeInteger|string`, for multiple reasons:
* allowing `a.succ("42")` seems pretty weird, if you want to create a `BigInt` from a string, be explicit about it.
* it's simpler to just allow `int`, as that is the common case and that is what `inc/dec` use
* `def` is a pretty meaningless name (i don't even know what it's supposed to mean), we should choose a good name, since argument names are part of the public API (named arguments)

Also, please make them `func`s, all the other functions already are `func`s.

konsumlamm:
@Araq may I ask why you need `succ`/`pred` rather than simply using `inc`/`dec` or `+`/`-` with `initBigInt`? Not that I'm opposed to adding those functions (they mirror the existing `succ`/`pred` for `Ordinal`s), but I don't see a reason why you'd "really need" them.

Araq:
I really need them in generic code. Alternatively you could offer a way to generate a literal like -1, 0, 1 of type T. For the builtin numbers the syntax is `T(1)` but for bigints it's `initBigInt(1)`, this is quite bad for generic code.

Here is an example from real code:

```nim

proc doesOverlapOrIncl*[T](s: var seq[(T, T)]; c, d: T): bool =
  mixin succ, pred
  assert c <= d
  var patched = false
  for i in 0..<s.len:
    let (x, y) = s[i]
    # X..Y and C..D overlap iff (X <= D and C <= Y)
    if x <= d and c <= y:
      s[i][0] = min(x, c)
      s[i][1] = max(y, d)
      return true
    elif succ(y) == c:
      s[i][1] = max(y, d)
      # but continue the search loop!
      patched = true
    elif pred(x) == d:
      s[i][0] = min(x, c)
      # but continue the search loop!
      patched = true
  if not patched:
    s.add (c, d)
  return false
```

Araq:
The PR is off, the second argument of pred/succ is really just `int`, much like `pointer +- offset` the type of the default parameter does not depend on the type of the first parameter.

Araq:
> def is a pretty meaningless name (i don't even know what it's supposed to mean)

It was short for `default`, but it's not a good name, I agree.

pietroppeter:
> we should choose a good name, since argument names are part of the public API (named arguments)

delta?

I am also curious about what is @Araq is doing with bigints. Using them in the compiler (as I remember you mentioned sometimes)?

pietroppeter:
or if we want to be consistent with stdlib the arguments will be `x` and `y`: https://nim-lang.org/docs/system.html#succ%2CT%2Cint

Araq:
> Using them in the compiler (as I remember you mentioned sometimes)?

Yes.

konsumlamm:
> These can be optimized by using `inc`/`dec`, which already take an integer argument, rather than unnecessarily creating a new `BigInt`.

Uh, I just saw that `inc`/`dec` only support `int32`, so we'd either have to change their signature or special case the case if the number is not in the range of `int32`.

dlesnoff:
First, I want to propose `increment` and `decrement` for the default values of succ/pred. `delta` is not explicit enough.
`offset` could be ok too, since it takes an `int` as a parameter.
It feels wrong to me to write something like `a.inc(-8)` and especially `a.dec(-16)` but …
Then comes the question of the type. I have added `SomeInteger|string` since these are the types required by `initBigInt` but I agree with you that `string` might be too much.

> Also, please make them funcs, all the other functions already are funcs.
Oops. Forgot that. I made a PR to change the procedures into functions but :)

Actually, I am personnaly strongly opposed to `pred` and `succ` functions. I think their names are too confusing for the user and are redundant with dec and inc. They might mirror the Ordinal type, but I do not see why BigInt functions should do this.

> Alternatively you could offer a way to generate a literal like -1, 0, 1 of type T. For the builtin numbers the syntax is T(1) but for bigints it's initBigInt(1)

I am more for this kind of change but I do not know how to fix this. Maybe the original motivation for such a long and specific name as `initBigInt` is to make the user aware that this allocates (potentially a lot) memory ?

Another problem is the addition with `int`. We should implement addition, subtraction and multiplication between a `BigInt` and an `int` in general too. This would avoid the need for type conversion for every operation.

> allowing a.succ("42") seems pretty weird, if you want to create a BigInt from a string, be explicit about it (you can always do a + initBigInt("42"))
Then why should we accept `a.inc(42)` rather than enforce `a + 42.initBigInt` ? For functional programming ?

konsumlamm:
> Actually, I am personnaly strongly opposed to `pred` and `succ` functions. I think their names are too confusing for the user and are redundant with dec and inc. They might mirror the Ordinal type, but I do not see why BigInt functions should do this.

As Araq said, so that you can use them in generic functions that also work on `Ordinal`s.

> > Alternatively you could offer a way to generate a literal like -1, 0, 1 of type T. For the builtin numbers the syntax is T(1) but for bigints it's initBigInt(1)
> 
> I am more for this kind of change but I do not know how to fix this. Maybe the original motivation for such a long and specific name as `initBigInt` is to make the user aware that this allocates (potentially a lot) memory ?

I don't think that's possible currently. If we just renamed `initBigInt` to `BigInt`, we'd get a redefinition error.

> Then why should we accept `a.inc(42)` rather than enforce `a + 42.initBigInt` ? For functional programming ?

Because `a.inc(42)` is more efficient than `a += 42.initBigInt`, since it doesn't allocate an intermediate `BigInt`. This has absolutely nothing to do with functional programming.

narimiran:
I've created https://github.com/nim-lang/bigints/pull/81 because @Araq needs this functionality ASAP.

This PR can refine and optimize it (use `inc`/`dec`, etc.) later.

Issue #59, opened 2022-01-01
JS backend


konsumlamm:
I experimented a bit with running the tests on the JS backend, but I came to the conclusion that this is currently not viable. The problem is https://github.com/nim-lang/Nim/issues/4714: 64 bit integers are broken on the JS backend, since they're implemented as floating point values.

We *could* use `std/jsbigints` on the JS backend, but I'd rather not do that, since it has a different API, which means we'd have to implement some functions twice. Apart from that, this library wouldn't be a pure Nim implementation anymore then.

konsumlamm:
A possible (temporary) fix for this would be to use `uint16`/`uint32` instead of `uint32`/`uint64` on the JS backend. That should only require changing a few types and constants, which should be doable by placing some type aliases and consts in a `when` at the top.

PR #82, opened 2022-01-21
Enable JS tests in VM if Nim version is >= 1.6

konsumlamm:
The tests are only run in the VM and only if the Nim version is >= 1.6, since unsigned integers are broken for JS otherwise (see #59). However, this has the advantage that we now effectively test a 32 bit target and it made me find & fix some bugs in the tests for `toSignedInt` that assumed 64 bit `int`s.

### DIFF
--- a/bigints.nimble
+++ b/bigints.nimble
@@ -12,7 +12,7 @@ srcDir      = "src"
 requires "nim >= 1.4.0"
 
 task test, "Test bigints":
-  for backend in ["c", "cpp"]:
+  for backend in ["c", "cpp", "js"]:
     echo "testing " & backend & " backend"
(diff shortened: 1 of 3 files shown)


dlesnoff (approved):

konsumlamm:
@narimiran what do you think?

narimiran:
I'd rather wait (or work on ;)) for the upstream fix (https://github.com/nim-lang/Nim/issues/4714) and for now not officially support JS, than to have these workarounds and semi-working state.

konsumlamm:
Ok, I'll move the test fixes to another PR then.

xflywind:
I will look into https://github.com/nim-lang/Nim/issues/4714 too.

PR #84, opened 2022-01-22
Add + and _ support for BigInt initialization, closes #52

### LINKED ISSUE #52: Support more characters for initializing a `BigInt` from a string
Analogous to `-` for negative numbers, we could also support a leading `+`, which would be ignored. Another thing that would be nice is supporting `_` (which also get ignored), akin to normal int literals:
```nim
let a = initBigInt("+10_000_000_000")
```

dlesnoff:
We can now initialize a BigInt with a number whose digits are separated by `'_'`. There is also a support for `'+'` symbol in front of the number. If the number has an incorrectly placed `'_'`, raises a `ValueError`.
Closes #52.
Right now, the code uses two more imports sequtils and strutils. sequtils import might be removed. In this case, we need to remove the line with filterIt that removes `_` and uses a while iteration. Drawback : we need to use a var statement instead of a let statement.


src/bigints.nim
```diff
@@ -910,9 +910,44 @@ func initBigInt*(str: string, base: range[2..36] = 10): BigInt =
...
+    discard
+  of 'a'..'z':
+    discard
+  of 'A'..'Z':
+    discard
+  else:
+    raise newException(ValueError, "The digit has not been recognized: " & str[0])
+  
+  # We check positions of '_' inside str
+  let posFirstSeparator = str[first..str.high].find('_')
```
dlesnoff, line 931: This find could be replaced by a for loop if we do not want to import strutils.

src/bigints.nim
```diff
@@ -910,9 +910,44 @@ func initBigInt*(str: string, base: range[2..36] = 10): BigInt =
...
+  of '+':
+    first = 1
+  of '_':
+    raise newException(ValueError, "A number can not begin with an '_'")
+  of '0'..'9':
+    discard
+  of 'a'..'z':
+    discard
+  of 'A'..'Z':
+    discard
```
dlesnoff, line 926: We could replace these three cases by:
```
of HexDigits:
  discard
```
if we agree to import strutils.

src/bigints.nim
```diff
@@ -910,9 +910,44 @@ func initBigInt*(str: string, base: range[2..36] = 10): BigInt =
...
+
+  if withSeparator:
+    for i in posFirstSeparator + 1 .. str.len - 1:
+      if str[i] == '_' and i mod 4 != (posFirstSeparator + first) mod 4:
+        raise newException(ValueError, "An underscore '_' has not been placed every three digits" & $i & ' ' & $posFirstSeparator & ' ' & $first)
+      if str[i] != '_' and i mod 4 == (posFirstSeparator + first) mod 4:
+        raise newException(ValueError, "There is a missing underscore '_' that should have been placed every three digits " & $i & ' ' & $posFirstSeparator)
+    if str[^1] == '_':
+      raise newException(ValueError, "Underscore should not be placed at the end of the string")
+  let str = str.filterIt(it != '_')
```
dlesnoff, line 950: Needs sequtils

tests/tbigints.nim
```diff
@@ -19,6 +19,7 @@ proc main() =
     let d = -123456789012345678.initBigInt
     doAssert $d == "-123456789012345678"
 
+
```
konsumlamm, line 22:
```suggestion
```
Why?

tests/tbigints.nim
```diff
@@ -31,6 +32,58 @@ proc main() =
     let h = 1234567.initBigInt
     let i = h.initBigInt
     doAssert $h == $i
+  
+    block:
```
konsumlamm, line 36: We should also test `"-".initBigInt` and `"+".initBigInt` (those currently fail I think).

tests/tbigints.nim
```diff
@@ -31,6 +32,58 @@ proc main() =
...
+
+    block:
+      # Wrong formatting of numbers should raise a ValueError
+      doAssertRaises(ValueError): discard "1_2_345_678".initBigInt
+      doAssertRaises(ValueError): discard "_123_345_678".initBigInt
+      doAssertRaises(ValueError): discard "_12_3_345_678".initBigInt
+      doAssertRaises(ValueError): discard "123_345_678_".initBigInt
+      doAssertRaises(ValueError): discard "123_345_67_8".initBigInt
+      doAssertRaises(ValueError): discard "12_33_45_67".initBigInt
+      doAssertRaises(ValueError): discard "123_345_678912".initBigInt
```
konsumlamm, line 85: I don't think e.g. `"12_33_45_67".initBigInt` should be disallowed. First of all, it's also allowed for normal integer literals (which I think should be a guiding design for this). It also has valid use cases, especially with binary/hexadecimal, when you want to group specific bits/bytes. And apart from that, it makes the implementation more complex, with basically no gain.

src/bigints.nim
```diff
@@ -910,9 +910,44 @@ func initBigInt*(str: string, base: range[2..36] = 10): BigInt =
...
 
-  if str[0] == '-':
+  case str[0]
+  of '-':
     first = 1
     neg = true
+  of '+':
+    first = 1
+  of '_':
+    raise newException(ValueError, "A number can not begin with an '_'")
```
konsumlamm, line 920:
```suggestion
    raise newException(ValueError, "A number can not begin with _")
```

src/bigints.nim
```diff
@@ -910,9 +910,44 @@ func initBigInt*(str: string, base: range[2..36] = 10): BigInt =
...
+  of '0'..'9':
+    discard
+  of 'a'..'z':
+    discard
+  of 'A'..'Z':
+    discard
+  else:
+    raise newException(ValueError, "The digit has not been recognized: " & str[0])
```
konsumlamm, line 928: These checks should be after checking for `-`/`+`, since now, e.g. `+_42` isn't catched by this.

src/bigints.nim
konsumlamm, line 950: We should just ignore any underscores, except when they're at the start or the end.

konsumlamm: That creates a new string, I think it would be better to just skip the underscores while parsing.

src/bigints.nim
dlesnoff: This is catched by the second case of statement, but +@42 isn’t catched by this indeed.

tests/tbigints.nim
dlesnoff: I added these tests, and strangely those not fail. I have to add two if-else clause !

tests/tbigints.nim
dlesnoff: I misunderstood the normal integer design. I will adapt my code.

dlesnoff:
I fear I will have to make a copy. Only way to avoid this would be either to check first for _ in the block considered in the accumulator, or to decrease size after making it immutable.

konsumlamm (commented):
I think you might be right that we need to filter the input string first in the case that `base` is not a power of 2, since the first inner for loop uses the fact that the number of digits is known to avoid a multiplication in each step.

However, it should be possible to do the filtering in-place, by moving the non-underscore blocks together, so that no additional string is allocated (remember, strings are mutable). If the string has no underscores, this should just be a linear traversal.

src/bigints.nim
```diff
@@ -954,9 +954,28 @@ func initBigInt*(str: string, base: range[2..36] = 10): BigInt =
...
+  of '_':
+    raise newException(ValueError, "A number can not begin with _")
+  of '0'..'9':
+    discard
+  of 'a'..'z':
+    discard
+  of 'A'..'Z':
+    discard
+  else:
+    raise newException(ValueError, "Unrecognized digit: " & str[0])
```
konsumlamm, line 976:
```suggestion
  else:
    discard
  if str[first] == '_':
    raise newException(ValueError, "A number can not begin with _")
```
Otherwise the `+_1` test fails.

src/bigints.nim
```diff
@@ -961,6 +961,16 @@ func parseDigit(c: char, base: uint32): uint32 {.inline.} =
   if result >= base:
     raise newException(ValueError, "Invalid input: " & c)
 
+func filterUnderscores(str: var string) =
```
konsumlamm, line 964:
```suggestion
func filterUnderscores(str: var string) {.inline.} =
```
This probably gets inlined anyway, but better be explicit about it.

dlesnoff:
The PR should now pass the tests.
For now, I make a copy only in latter case and used your function to do the in-place filtering of underscores.
I think there is some variable (the accumulator, I do not know anymore what I meant in my last comment ?) whose value can be decreased, to not output a zero, when an underscore is met.

src/bigints.nim
```diff
@@ -989,33 +1012,38 @@ func initBigInt*(str: string, base: range[2..36] = 10): BigInt =
...
     result.normalize()
   else:
+    var str = str
+    filterUnderscores(str)
     let d = initBigInt(base ^ size)
     for i in countup(first, str.high, size):
       var num = 0'u32 # the accumulator in this block
       if i + size <= str.len:
         # iterator over a block of length `size`, so we can use `d`
         for j in countup(i, min(i + size - 1, str.high)):
```
konsumlamm, line 1034:
```suggestion
        for j in countup(i, i + size - 1):
```
I just noticed that the `min` is unnecessary here, since we check `i + size <= str.len`.

konsumlamm:
The issue is that `unsignedAdditionInt(result, result * d, num)` relies on the `for`-loop before running exactly `size` times (otherwise we couldn't use the precomputed `d`). We could just always run the other `for`-loop, but that computes a new `BigInt` (`mul`) instead. A way to avoid extra allocations would be to count the underscores to iterate over exactly `size` non-underscore chars. However, I think that can be done in a future PR, if it is deemed necessary.

konsumlamm (approved):

konsumlamm:
@narimiran what do you think?

dlesnoff:
@narimiran Can we merge this PR ? Still no opinion about our question ?

PR #85, opened 2022-01-22
Make `inc`/`dec` take an `int`

konsumlamm:
Also add doc comments for `succ`/`pred`. Supersedes #79.


dlesnoff:
You should add tests for `a.inc(int32.high+1)`, `a.inc(int64.high)`, and same for dec and int64.low, int32.low-1.

konsumlamm:
@dlesnoff done.

The CI failure is caused by #88 (unrelated to this PR). See #89 for a (temporary) fix.

PR #89, opened 2022-01-23
Fix the arc & orc GC related invmod issue closes #88

### LINKED ISSUE #88: Tests of invmod fail with arc & orc GC
The latest commit does not pass due to to an error with the arc garbage collector and the modular inverse function invmod.
It fails at the very first test of the function line 388.

How can I debug the invmod function to fix it ?

dlesnoff:
Closes #88.
Thanks to @konsumlamm

### DIFF
--- a/src/bigints.nim
+++ b/src/bigints.nim
@@ -1041,7 +1041,7 @@ func invmod*(a, modulus: BigInt): BigInt =
     while r1 > 0:
       let
         q = r0 div r1
-        # The ternary condition is needed for arc and orc GC
+        # the `q.isZero` check is needed because of an ARC/ORC bug (see https://github.com/nim-lang/bigints/issues/88)
         rk = if q.isZero: r0 else: r0 - q * r1
         sk = if q.isZero: s0 else: s0 - q * s1
       r0 = r1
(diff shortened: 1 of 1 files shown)


konsumlamm:
Eh, I'm not really happy with this worksround, I'd rather find out what the bug is and fix it.

src/bigints.nim
```diff
@@ -1041,8 +1041,9 @@ func invmod*(a, modulus: BigInt): BigInt =
     while r1 > 0:
       let
         q = r0 div r1
-        rk = r0 - q * r1
-        sk = s0 - q * s1
+        # The ternary condition is needed for arc and orc GC
+        rk = if q.isZero: r0 else: r0 - q * r1
+        sk = if q.isZero: s0 else: s0 - q * s1
```
konsumlamm, line 1046:
```suggestion
        # the `q.isZero` check is needed because of an ARC/ORC bug (see https://github.com/nim-lang/bigints/issues/88)
        rk = if q.isZero: r0 else: r0 - q * r1
        sk = if q.isZero: s0 else: s0 - q * s1
```

konsumlamm (approved):
I guess the fix is fine to fix the CI. In the long run, we might want to replace the algorithm by the extended binary GCD algorithm anyway (I'm currently testing if that's more efficient).

PR #93, opened 2022-01-26
Optimize `((a mod modulus) + modulus) mod modulo`

konsumlamm:
Avoids doing `mod` twice.

I also swapped `a.modulo(modulus)` and `modulus` in `invmod`, to skip the first iteration (otherwise `q` would always be zero in the first iteration, since `a.modulo(modulus) < modulus`).

### DIFF
--- a/src/bigints.nim
+++ b/src/bigints.nim
@@ -1068,6 +1068,14 @@ iterator `..<`*(a, b: BigInt): BigInt =
     yield res
     inc res
 
+
+func modulo(a, modulus: BigInt): BigInt =
+  ## Like `mod`, but the result is always in the range `[0, modulus-1]`.
+  ## `modulus` should be greater than zero.
+  result = a mod modulus
+  if result < 0:
+    result += modulus
+
 func invmod*(a, modulus: BigInt): BigInt =
   ## Compute the modular inverse of `a` modulo `modulus`.
   ## The return value is always in the range `[1, modulus-1]`
@@ -1083,23 +1091,23 @@ func invmod*(a, modulus: BigInt): BigInt =
     raise newException(DivByZeroDefect, "0 has no modular inverse")
   else:
     var
-      r0 = ((a mod modulus) + modulus) mod modulus
-      r1 = modulus
-      s0 = one
-      s1 = zero
+      r0 = modulus
+      r1 = a.modulo(modulus)
+      t0 = zero
+      t1 = one
     while r1 > 0:
       let
         q = r0 div r1
         # the `q.isZero` check is needed because of an ARC/ORC bug (see https://github.com/nim-lang/bigints/issues/88)
         rk = if q.isZero: r0 else: r0 - q * r1
-        sk = if q.isZero: s0 else: s0 - q * s1
+        tk = if q.isZero: t0 else: t0 - q * t1
       r0 = r1
       r1 = rk
-      s0 = s1
-      s1 = sk
+      t0 = t1
+      t1 = tk
     if r0 != one:
       raise newException(ValueError, $a & " has no modular inverse modulo " & $modulus)
-    result = ((s0 mod modulus) + modulus) mod modulus
+    result = t0.modulo(modulus)
 
 func powmod*(base, exponent, modulus: BigInt): BigInt =
   ## Compute modular exponentation of `base` with power `exponent` modulo `modulus`.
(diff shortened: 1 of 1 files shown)


konsumlamm:
Hmm, looks like there's a Nim 1.4 bug...

src/bigints.nim
```diff
@@ -1119,8 +1127,7 @@ func powmod*(base, exponent, modulus: BigInt): BigInt =
     if exponent < 0:
       base = invmod(base, modulus)
       exponent = -exponent
-    var
-      basePow = ((base mod modulus) + modulus) mod modulus # Base stays in [0, m-1]
+    var basePow = base.modulo(modulus) # base stays in [0, modulus-1]
```
dlesnoff, line 1130: Please, remove the comment. The function should be explicit enough.

src/bigints.nim
```diff
@@ -1068,6 +1068,14 @@ iterator `..<`*(a, b: BigInt): BigInt =
     yield res
     inc res
 
+
+func modulo(a, modulus: BigInt): BigInt =
+  ## Like `mod`, but the result is always in the range `[0, modulus-1]`.
+  ## `modulus` should be greater than zero.
+  result = a mod modulus
+  if result < 0:
+    result += modulus
```
dlesnoff, line 1077: Maybe `positiveModulo` ? `canonicalModulo` ? I think the name of the function can be a little more explicit.

konsumlamm: It's the mathematical modulo operation, which is called just "modulo". The name doesn't really matter though, this function isn't exported anyway.

konsumlamm:
I managed to work around the bug, but I have no idea what caused it.

src/bigints.nim
```diff
@@ -1134,3 +1133,5 @@ func powmod*(base, exponent, modulus: BigInt): BigInt =
         result = (result * basePow) mod modulus
       basePow = (basePow * basePow) mod modulus
       exponent = exponent shr 1
+
+if false: discard invmod(zero, one)
```
dlesnoff, line 1137: This is strange. An if false never executes !

src/bigints.nim
```diff
@@ -1095,12 +1095,11 @@ func invmod*(a, modulus: BigInt): BigInt =
...
+    var rk, tk: BigInt # otherwise t1 is incorrectly inferred as cursor (https://github.com/nim-lang/Nim/issues/19457)
     while r1 > 0:
-      let
-        q = r0 div r1
-        # the `q.isZero` check is needed because of an ARC/ORC bug (see https://github.com/nim-lang/bigints/issues/88)
-        rk = if q.isZero: r0 else: r0 - q * r1
-        tk = if q.isZero: t0 else: t0 - q * t1
+      let q = r0 div r1
+      rk = r0 - q * r1
+      tk = t0 - q * t1
```
dlesnoff, line 1102: So you remove a workaround for another workaround in mod function ?
Isn't this PR only for replacement of ((a mod modulus) + modulus) mod modulo ?

src/bigints.nim
konsumlamm: Oops, that's just a debugging leftover.

src/bigints.nim
konsumlamm: This new workaround works for for both #88 and whatever caused the 1.4 failure. The latter was introduced by the change from `((a mod modulus) + modulus) mod modulus` to `a.modulo(modulus)`.

src/bigints.nim
```diff
@@ -1083,23 +1091,22 @@ func invmod*(a, modulus: BigInt): BigInt =
...
     var
-      r0 = ((a mod modulus) + modulus) mod modulus
-      r1 = modulus
-      s0 = one
-      s1 = zero
+      r0 = modulus
+      r1 = a.modulo(modulus)
+      t0 = zero
+      t1 = one
+    var rk, tk: BigInt # otherwise t1 is incorrectly inferred as cursor (https://github.com/nim-lang/Nim/issues/19457)
```
dlesnoff, line 1098: Why did you change s0 and s1 to t0 and t1 ? You said in my PR #65 that s was the variable name used pretty much anywhere else and made a point to change the variable name then to s.
I do not understand why you swaped values of r0 and r1 too (and t0 and t1 accordingly). Is it necessary to fix the code ? Isn't it rather to speed up the first step ? 
since `a.modulo(modulus)` is already reduced, q will not reduce anything at first step.

konsumlamm: As I said in https://github.com/nim-lang/bigints/pull/93#issue-1115242625, I only swapped `r0` and `r1` to avoid one iteration (before, in the first iteration, `r0` was always smaller than `r1`, so `r0 div r1` was always zero).

As for changing s to t: The standard naming scheme is that we compute `s` and `t` such that `a * s + b * t = gcd(a, b)`. Originally, we had a as `a.modulo(modulus)` and b as `modulus`, so if the GCD is 1, the modular inverse of a would be s. However, now with switching a and b, we have a as `modulus` and b as `a.modulo(modulus)`, so that the modular inverse is now t.

This is not strictly related to optimizing `((a mod modulus) + modulus) mod modulo`, but it's another optimization to `invmod`, so I don't see a reason not to include it in this PR.

dlesnoff (approved):
These changes seems fine to me.

PR #95, opened 2022-01-29
Karatsuba multiplication

dlesnoff:
This is still a draft.
I get a strange error in one test of the tliterals.nim file. I have no clue why this test is wrong when I switch the karatsuba multiplication.


dlesnoff:
Karatsuba multiplication now works successfully. It remains to find the optimal treshold at which we switch the multiplication mode.

dlesnoff:
Lack tests for multiplication.

src/bigints.nim
```diff
@@ -418,6 +418,26 @@ func unsignedMultiplication(a: var BigInt, b, c: BigInt) {.inline.} =
...
+  a.limbs.setLen(bl + 1)
+  var tmp = 0'u64
+
+  for i in 0 ..< bl:
+    tmp += uint64(b.limbs[i]) * uint64(c)
+    a.limbs[i] = uint32(tmp and uint32.high)
+    tmp = tmp shr 32
+
+  a.limbs[bl] = uint32(tmp)
+  normalize(a)
```
konsumlamm, line 434:
```suggestion
func scalarMultiplication(a: var BigInt, b: BigInt, c: uint32) {.inline.} =
  if c == 0:
    return zero
  let
    bl = b.limbs.len
  a.limbs.setLen(bl + 1)
  var tmp = 0'u64
  let c = uint64(c)

  for i in 0 ..< bl:
    tmp += uint64(b.limbs[i]) * c
    a.limbs[i] = uint32(tmp and uint32.high)
    tmp = tmp shr 32

  a.limbs[bl] = uint32(tmp)
  normalize(a)
```

src/bigints.nim
```diff
@@ -418,6 +418,26 @@ func unsignedMultiplication(a: var BigInt, b, c: BigInt) {.inline.} =
...
+
+# forward declaration for use in `multiplication`
+func karatsubaMultiplication(a: var BigInt, b, c: BigInt) {.inline.}
```
konsumlamm, line 437: `karatsubaMultiplication` is recursive, it makes no sense to inline that.

src/bigints.nim
```diff
@@ -428,11 +448,69 @@ func multiplication(a: var BigInt, b, c: BigInt) =
     cl = c.limbs.len
 
   if cl > bl:
-    unsignedMultiplication(a, c, b)
+    if bl <= karatsubaTreshold:
+      karatsubaMultiplication(a, c, b)
+    else:
+      unsignedMultiplication(a, c, b)
```
konsumlamm, line 454:
```suggestion
    if bl > karatsubaTreshold:
      karatsubaMultiplication(a, c, b)
    else:
      unsignedMultiplication(a, c, b)
```

src/bigints.nim
```diff
@@ -428,11 +448,69 @@ func multiplication(a: var BigInt, b, c: BigInt) =
...
   else:
-    unsignedMultiplication(a, b, c)
+    if cl <= karatsubaTreshold:
+      karatsubaMultiplication(a, b, c)
+    else:
+      unsignedMultiplication(a, b, c)
```
konsumlamm, line 459:
```suggestion
    if cl > karatsubaTreshold:
      karatsubaMultiplication(a, b, c)
    else:
      unsignedMultiplication(a, b, c)
```

konsumlamm (commented):
Note: I haven't looked at `karatsubaMultiplication` itself yet.

src/bigints.nim
```diff
@@ -428,11 +448,69 @@ func multiplication(a: var BigInt, b, c: BigInt) =
...
+    else:
+      unsignedMultiplication(a, c, b)
+    return
+  if cl < karatsubaTreshold:
+    if bl <= cl:
+      unsignedMultiplication(a, c, b)
+    else:
+      unsignedMultiplication(a, b, c)
+    return
```
konsumlamm, line 485: I think it makes more sense to move this logic into `unsignedMultiplication`. Then we avoid checking this twice when calling `karatsubaMultiplication` from `multiplication` and we can call `unsignedMultiplication` inside `karatsubaMultiplication` (instead of `karatsubaMultiplication` itself).

src/bigints.nim
```diff
@@ -428,11 +448,69 @@ func multiplication(a: var BigInt, b, c: BigInt) =
...
+  let k = n shr 1 # should it be ceil(n/2) ?
+  var
+    low_b, high_b, low_c, high_c: BigInt
+  # Decompose `b` and `c` in two parts of (almost) equal length
+  low_b.limbs = b.limbs[0 .. k-1]
+  high_b.limbs = b.limbs[k .. ^1]
+  low_c.limbs = c.limbs[0 .. k-1]
+  high_c.limbs = c.limbs[k .. ^1]
```
konsumlamm, line 493: These all create new `seq`s, which isn't very efficient. Perhaps we should use `openArray[uint32]` instead (then this can use `toOpenArray` for O(1) slicing) or make our own "sliceable seq" to avoid copies.

src/bigints.nim
```diff
@@ -428,11 +448,69 @@ func multiplication(a: var BigInt, b, c: BigInt) =
...
+  
+  # subtractive version of Karatsuba's algorithm :
+  # limit carry handling in opposition to the additive version
+  var
+    lowProduct, highProduct, A3, A4, A5, middleTerm: BigInt = zero
+  karatsubaMultiplication(lowProduct, low_b, low_c)
```
konsumlamm, line 499: `low_b` and `low_c` aren't necessarily normalized afaict, so that may cause problems.

src/bigints.nim
```diff
@@ -428,11 +448,69 @@ func multiplication(a: var BigInt, b, c: BigInt) =
...
+  karatsubaMultiplication(highProduct, high_b, high_c)
+  A3 = low_b - high_b # Additive variant of Karatsuba
+  A4 = low_c - high_c # would add them
```
konsumlamm, line 502: Can we please avoid uppercase variable names?

src/bigints.nim
dlesnoff: I would like to but we can not. Karatsuba has to be signed. Intermediary results are necessary signed, even if it multiplies only unsigned integers.
I have fixed many bugs in the version you just reviewed, I am sorry I should have warned you, there will be many changes to this version.
First, I think I will make karatsubaMultiplication and multiplication completely independent as for now. I need to add many tests, and I can not call unsignedMultiplication, for small cases yet.

src/bigints.nim
dlesnoff: Ok, I did not thougt about this. I used seq because I expect to join the results into a seq after.

src/bigints.nim
dlesnoff: No problem. I am going to change this asap.

src/bigints.nim
dlesnoff: Removed.

src/bigints.nim
```diff
@@ -428,11 +448,69 @@ func multiplication(a: var BigInt, b, c: BigInt) =
...
+    bl = b.limbs.len
+    cl = c.limbs.len
+  let n = max(bl, cl)
+  if bl == 1:
+    # base case : multiply the only limb with each limb of second term
+    scalarMultiplication(a, c, b.limbs[0])
+    return 
+  if cl == 1:
+    scalarMultiplication(a, b, c.limbs[0])
+    return
```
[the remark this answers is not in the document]
dlesnoff: The idea was to reduce the number of operations if we know there is only one limb.
We do not have to do a whole for loop.
But you are right, we can do a scalarMultiplication with unsignedMultiplication.

src/bigints.nim
dlesnoff: This might be the problem I encounter at compilation time. I have not done any test when the operands does not have the same limbs sequence’s length.

src/bigints.nim
konsumlamm: It's fine, I didn't expect this to be the final version anyway. As I said, I haven't looked at `karatsubaMultiplication` in depth yet, since I want to better understand the algorithm first.

src/bigints.nim
dlesnoff: @konsumlamm openArray are only used for procs arguments, when we want either a seq or an array as argument.
Can't we use some pointers here ? Do we have to create a new structure ?
Internal structure is a seq, if we use another structure, we would have to make a copy anyway or change the internal structure.
We can not use arrays neither, since we do not know the size at compile time. (It depends on the variable k, which value is known at runtime).

konsumlamm: An `openArray` is basically a pointer and a length, it doesn't create a copy. Anything that doesn't create a copy but just modifies indices/pointers should be good.

dlesnoff: I do not see how we can call multiplication then on those pointers, without making a multiplication directly on arrays.
We need to initialize BigInts

konsumlamm: That can be solved by `karatsubaMultiplication` taking an `openArray` (or some custom slice type) instead of a `BigInt` (it doesn't need the sign anyway).

demotomohiro (changes requested):

src/bigints.nim
```diff
@@ -430,11 +453,73 @@ func multiplication(a: var BigInt, b, c: BigInt) =
...
+    else:
+      unsignedMultiplication(a, b, c)
+    a.isNegative = b.isNegative xor c.isNegative
+    return
+  var
+    low_b, high_b, low_c, high_c: BigInt
+  # Decompose `b` and `c` in two parts of (almost) equal length
+  low_b.limbs = b.limbs[0 .. k-1]
+  high_b.limbs = b.limbs[k .. ^1]
+  low_c.limbs = c.limbs[0 .. k-1]
```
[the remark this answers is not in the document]
dlesnoff: I can simply change n to the min of bl and cl in this case.
I want to add these tests, but I am waiting for my other PRs to be merged. In these, I have added random generation and benchmarks. I am especially waiting for #112 . With initRandomBigInt, I will be able to generate bigints of a specific size for tests.

dlesnoff:
Thanks. I am not sure I can continue working on this PR, I am terribly busy.
I will try to move away the tests under the when isMainModule instruction
block.
Maybe I will test your suggestion, but I do not think I will have time to
do much debugging.

Furthermore, I am really concern by the choice of uint32 vs uint64 (as
suggested by @mratsim).
This seems to be a big change for the library, that we need to tackle
beforehand.

konsumlamm:
> Furthermore, I am really concern by the choice of uint32 vs uint64 (as suggested by @mratsim). This seems to be a big change for the library, that we need to tackle beforehand.

Changing that requires proper support for addition and multiplication of `uint64` without overflow (and I don't think writing that in assembly is a good solution). I don't see what that has to do with Karatsuba multiplication though, it doesn't really change the algorithm.

src/bigints.nim
dlesnoff: > These all create new `seq`s, which isn't very efficient. Perhaps we should use `openArray[uint32]` instead (then this can use `toOpenArray` for O(1) slicing) or make our own "sliceable seq" to avoid copies.

The algorithm wants the value of the polynomial corresponding to each parts of the slice.
The best way so far that I conceive is to modify the BigInt's limbs field from:
```nim
type
  BigInt* = object
    limbs: seq[uint32]
    isNegative: bool
 ```
to
```nim
type
  BigInt* = object
    limbs: ref seq[uint32]
    isNegative: bool
```
Otherwise, we will have to reimplement addition, subtraction and base case multiplication for another container.

mratsim: seqs are openarrays, if you implement for openarrays, it's implemented for seq and arrays.

dlesnoff: We have to change the whole code.

Addition, subtraction and multiplication take bigints with a sequence parameter as input.

For the Karatsuba algorithm as well as some others, we need to manipulate these sequences through a pointer and get a pointer for parts of the sequence.

We also need to operate on those slices of the sequence, i.e. get the value associated with each part of the slice, and add, subtract, and multiply these slices.

That's why we need a seq-like container with the possibility of getting a reference to each value of the seq.

None of the openarray types enables this.

demotomohiro: As @konsumlamm said, I think openArray and `toOpenArray` is good enough to implement karatsuba multiplication.
You can write recursive procedure like this using openArray and `toOpenArray`:
```nim
proc sum(x: openArray[int]): int =
  case x.len:
  of 0:
    0
  of 1:
    x[0]
  of 2:
    x[0] + x[1]
  else:
    let mid = x.len div 2
    sum(toOpenArray(x, 0, mid - 1)) + sum(toOpenArray(x, mid, x.high))

echo sum([1, 2, 3, 4, 5])
```

This is a part of `karatsubaMultiplication` in your current PR:
```nim
var
  low_b, high_b, low_c, high_c: BigInt
# Decompose `b` and `c` in two parts of (almost) equal length
low_b.limbs = b.limbs[0 .. k-1]
high_b.limbs = b.limbs[k .. ^1]
low_c.limbs = c.limbs[0 .. k-1]
high_c.limbs = c.limbs[k .. ^1]

# subtractive version of Karatsuba's algorithm to limit carry handling
var lowProduct, highProduct, add3, add4, add5, middleTerm: BigInt = zero

multiplication(lowProduct, low_b, low_c)
multiplication(highProduct, high_b, high_c)

add3 = low_b - high_b
add4 = high_c - low_c
```
Above code would be written using `toOpenArray` like:
```nim
# Decompose `b` and `c` in two parts of (almost) equal length
template low_b = toOpenArray(b.limbs, 0, k - 1)
template high_b = toOpenArray(b.limbs, k, b.limbs.high)
template low_c = toOpenArray(c.limbs, 0, k - 1)
template high_c = toOpenArray(c.limbs, k, c.limbs.high)

# subtractive version of Karatsuba's algorithm to limit carry handling
var lowProduct, highProduct, add3, add4, add5, middleTerm: BigInt = zero

multiplication(lowProduct, low_b, low_c)
multiplication(highProduct, high_b, high_c)

# This code requires subtraction proc that takes 2 `openArray`
add3 = low_b - high_b
add4 = high_c - low_c
```

src/bigints.nim
```diff
@@ -420,6 +419,30 @@ func unsignedMultiplication(a: var BigInt, b, c: BigInt) {.inline.} =
       inc pos
   normalize(a)
 
+func scalarMultiplication(a: var BigInt, b: BigInt, c: uint32) {.inline.} =
+  # always called with bl >= cl
```
dlesnoff, line 423: This comment needs to be removed, since `cl == 1` in this case.

src/bigints.nim
dlesnoff: Thank you for the time taken and the detailed changes.

```nim
# This code requires subtraction proc that takes 2 `openArray`
```
I just want to point out that scalarMultiplication and unsignedMultiplication will also have to take openArrays:
```nim
  if bl == 1:
    scalarMultiplication(a, c, b.limbs[0]) # b and c are openArrays
    a.isNegative = b.isNegative xor c.isNegative
    return 
 ...
  if bl < karatsubaThreshold:
    if cl <= bl:
      unsignedMultiplication(a, b, c) # b and c are openArrays here
    else:
      unsignedMultiplication(a, c, b) # same
    a.isNegative = b.isNegative xor c.isNegative
    return
   ...
```
I can make a `multiplication(a: var Bigint, b, c: openArray[uint32]) =` proc, and avoid to rewrite addition and shr for openArrays.

I read the unsignedMultiplication and scalarMultiplication proc algorithms again and they effectively don't need parameters to be BigInts.
The `substract` proc will be quite delicate to convert for OpenArray parameters though.

I will look into it.

PR #97, opened 2022-01-31
Update CI to check 1.6.4

konsumlamm:


dlesnoff:
I am pretty sure `stable` version is already `1.6.2`.
I do not see why we should remove `1.6.0` support.

konsumlamm:
True, stable is currently 1.6.2, so I guess checking 1.6 at all (or checking stable) is a bit redundant.

However, 1.6.2 is 1.6.0 with bug fixes, I don't see why we should keep 1.6.0 support.

dlesnoff:
You are right, we should remove 1.6.0 support.

narimiran:
Btw, 1.6.4 is around the corner.

konsumlamm:
> Btw, 1.6.4 is around the corner.

So should I rather wait for that?

narimiran:
> So should I rather wait for that?

Yeah, keep this open and once the 1.6.4 is out, we'll update and merge it.

konsumlamm:
@narimiran since 1.6.4 is out now, I updated the PR. I also removed `stable`, since that's 1.6.4 anyway and this way we can explicitly state the supported versions.

dlesnoff (approved):

PR #100, opened 2022-02-02
Rename `toSignedInt` to `toInt` & add support for unsigned integers

### LINKED ISSUE #99: [Bug] `toSignedInt[int64]` produces wrong results in certain cases
Example:
```nim
import bigints

echo toSignedInt[int64](-initBigInt(0xFFFFFFFF_00000000'u64))
```
This currently prints `some(-9223372036854775808)`, but the expected result is `none(int64)`.

The reason is the following lines:
https://github.com/nim-lang/bigints/blob/18e389928c2875d86e3697cd5997612497e2a2df/src/bigints.nim#L817-L819
This is only correct if `x.limbs[1] == uint32.high shr 1 + 1`, otherwise `x` is out of range.

konsumlamm:
Fixes #99.

Followup to #74. `toInt` also works with unsigned integers. I also cleaned up the implementation a bit and added a few more tests.
I could also change the name to something like `toInteger`, if that's preferred.

cc @Araq


src/bigints.nim
```diff
@@ -797,54 +797,78 @@ func gcd*(a, b: BigInt): BigInt =
...
 
 
-func toSignedInt*[T: SomeSignedInt](x: BigInt): Option[T] =
-  ## Converts a `BigInt` number to signed integer, if possible.
-  ## If the `BigInt` doesn't fit in a `T`', returns `none`;
-  ## otherwise returns `some(T)`.
+func toInt*[T: SomeInteger](x: BigInt): Option[T] =
+  ## Converts a `BigInt` number to an integer, if possible.
+  ## If the `BigInt` doesn't fit in a `T`, returns `none(T)`;
+  ## otherwise returns `some(x)`.
```
dlesnoff, line 803: Since you changed the name, can you modify the pidigits example accordingly ?
It uses toSignedInt and would be broken by this change.

dlesnoff, line 803: I am not sure we should reexplain the concept of an Option in the docstring. It does explain the use of a generic type though, so I am ok with this.

konsumlamm: Good call, we should really add the examples to the CI.

konsumlamm: It doesn't really reexplain anything, it just says what is returned when.

tests/tbigints.nim
```diff
@@ -595,68 +608,104 @@ proc main() =
...
+        c = initBigInt(uint64(u32h) + 1)
+      doAssert toInt[uint8](a) == none(uint8)
+      doAssert toInt[uint32](a) == some(u32h)
+      doAssert toInt[uint](a) == some(u32h.uint)
+      doAssert toInt[uint8](b) == none(uint8)
+      doAssert toInt[uint32](b) == some(u32h - 1)
+      doAssert toInt[uint](b) == some(u32h.uint - 1)
+      doAssert toInt[uint8](c) == none(uint8)
+      doAssert toInt[uint32](c) == none(uint32)
+      doAssert toInt[uint64](c) == some(u32h.uint64 + 1)
```
dlesnoff, line 663: I see no uint16 tests, could you add some ?

tests/tbigints.nim
```diff
@@ -595,68 +608,104 @@ proc main() =
...
+      doAssert toInt[int8](n) == none(int8)
+      doAssert toInt[int32](n) == none(int32)
+      doAssert toInt[int64](n) == none(int64)
+
+    block: # uint64
+      let
+        u64h = uint64.high
+        a = initBigInt(u64h)
+        b = initBigInt(u64h - 1)
+        c = initBigInt("18446744073709551616") # uint64.high + 1
```
dlesnoff, line 699: Why not `initBigInt(uint64.high) + 1.initBigInt` ?

tests/tbigints.nim
konsumlamm: Sure, the reason I didn't add them is that there are no `int16` tests either.

Araq:
Fine with me, I will update my code.

PR #103, opened 2022-04-03
Update elliptic.nim

planetis-m:
The most significant change is turning let into const - a huge amount of temporaries goes away. Other is porting from strutils to strformat. And using 'bi when possible

### DIFF
--- a/examples/elliptic.nim
+++ b/examples/elliptic.nim
@@ -1,24 +1,24 @@
 # By Cyther606: https://forum.nim-lang.org/t/522
 # Adapted from: https://github.com/wobine/blackboard101/blob/master/EllipticCurvesPart4-PrivateKeyToPublicKey.py
 import bigints
-import std/[math, strutils]
+import std/[math, strformat]
 
 const
-  one = 1.initBigInt
-  two = 2.initBigInt
-  zero = 0.initBigInt
+  one = initBigInt(1)
+  two = initBigInt(2)
+  zero = initBigInt(0)
 
 proc `^`(base: int; exp: int): BigInt = pow(base.initBigInt, exp)
 
 # Specs of the Bitcoin's curve - secp256k1
-let
+const
   primeCurve: BigInt = 2^256 - 2^32 - 2^9 - 2^8 - 2^7 - 2^6 - 2^4 - one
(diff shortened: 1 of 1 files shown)


planetis-m:
Error: undeclared identifier: 'bi'

narimiran:
> Error: undeclared identifier: 'bi'

That's why we need https://github.com/nim-lang/bigints/blob/master/src/bigints/private/literals.nim, for example. See comment there: "It is needed as a workaround for Nim's parser for versions <= 1.4."

Long story short: please revert those `'bi` changes.

dlesnoff:
So we have to drop support for versions <= 1.4 to define const litterals with Bigints ?
I appreciate the change from strutils to strformat, the echo looks nicer that way.

konsumlamm:
> So we have to drop support for versions <= 1.4 to define const litterals with Bigints ?

We have to drop support for Nim <= 1.4 when we want to use `'bi`.

konsumlamm (approved):

examples/elliptic.nim
```diff
@@ -50,20 +51,21 @@ proc ecMultiply(genPoint: tuple, scalarHex: BigInt): (BigInt, BigInt) =
...
+
+the uncompressed public key (not address):
+{publicKey}
+
+the uncompressed public key (HEX):
+04{publicKey[0].toString(base = 16):0>64}{publicKey[1].toString(base = 16):0>64}
+
+the official Public Key - compressed:
+{(if publicKey[1] mod two == one: "03" & publicKey[0].toString(base = 16)
+  else: "02" & publicKey[0].toString(base = 16)):0>64}
```
konsumlamm, line 68: The reason the CI is failing is that this doesn't work in Nim 1.4.8, I think we'll have to use a variable instead.

dlesnoff:
I don’t think we will remove 1.4.8 support. It has been released on 26th May and I believe we keep support for Nim versions that has been released less than one year before (or at least two Nim versions).
Otherwise, I do not understand how this impact your personal projects : we do support 1.6.
Is it really a needed feature here ?
I think you can simply move the code between brackets into a `let` statement, before the echo.

planetis-m:
No I meant various PRs I made in Nim are postponed because of build failures. Like https://github.com/nim-lang/Nim/pull/19360 https://github.com/nim-lang/threading/pull/6

PR #104, opened 2022-05-16
Always normalize

### LINKED ISSUE #44: Are `BigInt`s always normalized?
It seems to me that `BigInt`s are always normalized (i.e. the highest limb is never 0, except if the number is 0), however, this isn't documented.

If this is the case, I think at least `isZero(n)` could be optimized to just check `n.limbs.len == 1 and n.limbs[0]` (it currently uses a loop).

konsumlamm:
Closes #44.

I've verified that with my changes the returned `BigInt`s are always normalized and documented the invariants. This also simplifies the `isZero` implementation.

### DIFF
--- a/src/bigints.nim
+++ b/src/bigints.nim
@@ -4,6 +4,10 @@ import std/[algorithm, bitops, math, options]
 
 type
   BigInt* = object
+    ## An arbitrary precision integer.
+    # Invariants for `a: BigInt`:
+    # * if `a` is non-zero: `a.limbs[a.limbs.high] != 0`
+    # * if `a` is zero: `a.limbs.len <= 1`
     limbs: seq[uint32]
     isNegative: bool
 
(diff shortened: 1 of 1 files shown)


src/bigints.nim
```diff
@@ -4,6 +4,10 @@ import std/[algorithm, bitops, math, options]
 
 type
   BigInt* = object
+    ## An arbitrary precision integer.
+    # Invariants for `a: BigInt`:
+    # * if `a` is non-zero: `a.limbs[a.limbs.high] != 0`
+    # * if `a` is zero: `a.limbs.len <= 1`
```
dlesnoff, line 10: This invariant is not specific to zero. a.limbs.len might be 1, if it is a small constant.

dlesnoff:
@konsumlamm 
How much does the BigInt normalization during initialisation step affect performances ?
I wonder if this one is necessary.
Have you tried to compare the examples before and after the change ?

PR #107, opened 2022-07-12
Add setBit, clearBit and testBit

demotomohiro:
This PR adds `setBit`, `clearBit` and `testBit` procs.
They are same to `setBit`, `clearBit` and `testBit` in `std/bitops`.


konsumlamm:
I like this, however I'd prefer if this could treat negative numbers as if they were represented in 2's complement, like in #111. For that, if the input is negative, set `isNegative` to `false`, decrement it (this is done by `invertIn` in #111, although it copies the `BigInt` first), then treat every bit as inverted (and the other bits that aren't stored as 1, so `clearBit` would then need to update the length, instead of `setBit`) and keep in mind that the updated bit also needs to be inverted. Then, at the end, increment, and set `isNegative` to `true` again (this is done by `invertOut` in #111).

PR #108, opened 2022-07-25
Add signum function

rotu:


konsumlamm:
Should this really allocate a new `BigInt`? I feel like this should just return an `int`, so that the user can choose if they want that extra allocation, by manually calling `initBigInt`. Do you have a concrete use casein mind for this, where it's more useful to have this return a `BigInt`?

rotu:
> Should this really allocate a new `BigInt`? I feel like this should just return an `int`, so that the user can choose if they want that extra allocation, by manually calling `initBigInt`. Do you have a concrete use casein mind for this, where it's more useful to have this return a `BigInt`?

The reason it returns a `BigInt` is for the invariant `x == abs(x) * sgn(x)`. I'd be okay with returning an `int` if `BigInt` arithmetic allowed implicit widening (like other numeric types do).

konsumlamm:
I think we should support arithmetic where one of the arguments is an `int` sooner or later.

rotu:
> I think we should support arithmetic where one of the arguments is an `int` sooner or later.

Any reason why `initBigInt(SomeSignedInt): BigInt` isn't an implicit conversion? That seems like the easy solution

rotu:
NVM - opened https://github.com/nim-lang/bigints/pull/109
Feel free to tell me there why this is a bad idea :-p

PR #111, opened 2022-07-28
Implement bitwise operations for negative numbers

### LINKED ISSUE #51: Bitwise operators for negative numbers
Currently, the bitwise operators (`and`, `or`, `xor`) only work for non-negative numbers. I suggest also supporting negative numbers, treating them as if they were represented using 2's complement (the standard representation for signed integers). This would also allow implementing `not x`, by simply doing `-x - 1`.

konsumlamm:
Fixes #51.

### DIFF
--- a/src/bigints.nim
+++ b/src/bigints.nim
@@ -540,7 +540,7 @@ func invertIn(a: BigInt): BigInt =
 
 func invertOut(a: var BigInt) =
   inc(a)
-  a.isNegative = false
+  a.isNegative = true
 
 func `not`*(a: BigInt): BigInt =
   ## Bitwise `not` for `BigInt`s.
--- a/src/bigints.nim
+++ b/src/bigints.nim
@@ -1304,8 +1304,3 @@ func powmod*(base, exponent, modulus: BigInt): BigInt =
         result = (result * basePow) mod modulus
       basePow = (basePow * basePow) mod modulus
       exponent = exponent shr 1
-
-# example
-let a = -1234.initBigInt
-let b = "-12345678901234567890".initBigInt
-echo (a and b)
--- a/src/bigints.nim
+++ b/src/bigints.nim
@@ -1,4 +1,6 @@
 ## Arbitrary precision integers.
+##
+## The bitwise operations behave as if negative numbers were represented in 2's complement.
 
 import std/[algorithm, bitops, math, options]
 
@@ -10,6 +12,12 @@ type
     # * if `a` is zero: `a.limbs.len <= 1`
     limbs: seq[uint32]
     isNegative: bool
+
+
+# forward declarations
+func succ*(a: BigInt, b: int = 1): BigInt
+func inc*(a: var BigInt, b: int = 1)
+func dec*(a: var BigInt, b: int = 1)
 
 
 func normalize(a: var BigInt) =
@@ -486,9 +494,6 @@ func `shl`*(x: BigInt, y: Natural): BigInt =
   if carry > 0:
     result.limbs.add(uint32(carry shr (32 - b)))
 
-# forward declaration for use in `shr`
-func dec*(a: var BigInt, b: int = 1)
-
 func `shr`*(x: BigInt, y: Natural): BigInt =
   ## Shifts a `BigInt` to the right (arithmetically).
   runnableExamples:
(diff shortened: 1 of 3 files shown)


src/bigints.nim
```diff
@@ -1,4 +1,6 @@
 ## Arbitrary precision integers.
+##
+## The bitwise operations behave as if negative numbers were represented in 2's complement.
```
dlesnoff, line 3: I wonder whether the choices we make in the library's implementation should not rather be added to Readme.
They are redundant information with the source code (though easier to read), and are interesting to the end-user, which will not look at source code.

src/bigints.nim
```diff
@@ -525,50 +530,151 @@ func `shr`*(x: BigInt, y: Natural): BigInt =
...
 
 func `and`*(a, b: BigInt): BigInt =
   ## Bitwise `and` for `BigInt`s.
-  assert (not a.isNegative) and (not b.isNegative)
-  bitwiseAnd(result, a, b)
+  if a.isNegative and not a.isZero:
+    if b.isNegative and not b.isZero:
+      # - and -
+      let a = invertIn(a)
+      let b = invertIn(b)
```
dlesnoff, line 557: a and b are negative. So we are not changing the isNegative field and we are only decrementing them.
We can avoid these calls to invertIn, and simply decrementing them.
It avoids one function call, one copy into result and one affectation

tests/tbigints.nim
```diff
@@ -31,7 +31,7 @@ proc main() =
     let h = 1234567.initBigInt
     let i = h.initBigInt
     doAssert $h == $i
-  
+
```
dlesnoff, line 34: Probably a whitespace on a blank line I added in a previous commit :'(. Sorry for this.

src/bigints.nim
```diff
@@ -525,50 +530,151 @@ func `shr`*(x: BigInt, y: Natural): BigInt =
...
+        result.limbs[i] = a.limbs[i]
+      for i in m ..< b.limbs.len:
+        result.limbs[i] = b.limbs[i]
+      invertOut(result)
+    else:
+      # - and +
+      let a = invertIn(a)
+      result = b
+      for i in 0 ..< min(a.limbs.len, b.limbs.len):
+        result.limbs[i] = (not a.limbs[i]) and b.limbs[i]
```
dlesnoff, line 572: Same as above.

dlesnoff (changes requested):
The test outputs look consistent with my tests in python3.
I would simply consider moving the implementation's choice to the README.md and remove some function calls.
Even though the algorithm is more explicit with these calls, it may add unuseful copies.

src/bigints.nim
konsumlamm: This is a doc comment, it will appear in the [generated documentation of this library](https://nimdocs.com/nim-lang/bigints/bigints.html), which I'd expect users to read. Imo the readme is more about general information on the library and it already mentions the bitwise operators, now without the note that they don't work for negative numbers.

tests/tbigints.nim
konsumlamm: A tip: You can configure your editor to remove trailing whitespace automatically when saving.

src/bigints.nim
konsumlamm: Wdym? The `isNegative` fields are changed from `true` to `false`, so that the decrementing is correct (we need the absolute value minus one). I think however that we can instead increment the number (and ignore the `isNegative` field). I highly doubt that `invertIn`/`invertOut` aren't already inlined (avoiding the function call and potential copy), but I should probably check that.

dlesnoff: I made a confusion between isNegative and isPositive.
Yes, that is exactly what I propose. We have already test their sign, so let's just decrement them. If we can even avoid the creation of the two temporaries (by integrating the +1 directly in the operations) it would be much better.
For 1024 limbs BigInts a and b, we copy 2048 limbs for temporaries in addition to the 1024 allocated for the result, just to do one `a and b`.

konsumlamm: They're inlined when adding an `{.inline.}` pragma, so I did that.

konsumlamm:
@narimiran I think this is ready to be merged now.

PR #113, opened 2022-08-14
Benchmarks

dlesnoff:
I benchmarked multiplication, division, random generation of bigints. Without flags, the operations are slow, with flags, the timings looks correct to me.
The code source quality of benchBigints.nim can be improved with templates to do less copy-paste in the future.
I have also changed some of the examples to import them (by exporting some procedures and using whenIsmainmodule) so that I can benchmark them.
I added some corresponding nimble tasks.
I use `-d:lto -d:danger --opt:speed` flags for benchmarks.
I consider using benchy for benchmarks is ok, it does not make it a dependency of the full library, it is only required for benchmarks.
This PR might be really useful for easy comparison of our changes to the library.

### DIFF
--- a/tests/tbigints.nim
+++ b/tests/tbigints.nim
@@ -1,5 +1,6 @@
 import bigints
 import std/options
+import random
 
 const
   zero = initBigInt(0)
@@ -786,6 +787,11 @@ proc main() =
     doAssert pred(a, 3) == initBigInt(4)
     doAssert succ(a, 3) == initBigInt(10)
 
+  # block:
+  #   randomize()
+  #   let a: BigInt = initRandomBigInt(33)
+  #   echo a
 
-static: main()
+
+#static: main()
 main()
--- a/benchmarks/benchBigints.nim
+++ b/benchmarks/benchBigints.nim
@@ -1,0 +1,61 @@
+import bigints
+import benchy
+import random
+
+block: # Bench random generation
+  randomize()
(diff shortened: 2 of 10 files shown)


konsumlamm (changes requested):
Generally looks good to me! I'm not really fond of `benchy`, since it only displays numbers in milliseconds, but I don't know of a better alternative. The benchmarking code has a lot of duplication, that could be easily reduced by extracting the benchmarking code into a proc/template. I think we should also bench `BigInt`s with less than 100 limbs, e.g. 10, 2, or 1, since I think those will be more common for casual uses of `BigInt`s.

tests/tbigints.nim
```diff
@@ -786,6 +787,11 @@ proc main() =
...
     doAssert succ(a, 3) == initBigInt(10)
 
+  # block:
+  #   randomize()
+  #   let a: BigInt = initRandomBigInt(33)
+  #   echo a
 
-static: main()
+
+#static: main()
```
dlesnoff, line 796: I have to remove these comments and let this file be.

pietroppeter:
in the past I did some benchmarking (addition and multiplication, inspired by a Js benchmarking suite) during a PR (then removed the code). They were done using library criterion and published results in a generated md file. Maybe it can turn out useful for this PR: https://github.com/nim-lang/bigints/pull/35#issuecomment-733159960

PR #120, opened 2022-12-08
fix int set construction

bung87:


dlesnoff:
Thanks I notice that we should bump the CI tests from 1.6.4 to 1.6.10 Nim version.

konsumlamm (approved):

PR #127, opened 2023-06-19
Rename tests/tliterals for automated testing

ehmry:


dlesnoff:
Please provide a description even if you use the Github CLI.

I do not get how this change enables «automated testing». We removed this file from the nimble task test for a good reason. All the files in the directory are prefixed with a t, to prevent naming conflicts/confusion (there can be only one bigints.nim file for example).

pietroppeter:
my guess about this is that the concern was about using command the predefined `nimble test` task: https://github.com/nim-lang/nimble#tests that would run all files in tests that start with `t`. This is not a concern for this repo since task `test` is overwritten.

dlesnoff:
Superseded by #128.

Issue #70, opened 2022-01-09
[Feature request] Add more examples : factorization, find modular square-root


dlesnoff:
To showcase what can be done with the library, as well to benchmark it, I propose to add examples for factorization with simple exponential algorithms like rho-pollard, p-1 or p+1.
We would need for these gcd and powmod algorithms currently in PR #65 and #66.

We could also improve/adapt the [algorithm published on Rosetta Code for Shanks-Tonelli ](http://rosettacode.org/wiki/Tonelli-Shanks_algorithm#Nim). Some examples could not be computed with only nim's integer native library and have a need for multi-precision arithmetic.

Too much examples might make the library size big, but they help to define a good API, to understand which feature are essential and which one needs optimization.

PR #129, opened 2023-06-29
feat: add rho-Pollard example

vil02:
This PR adds an example of factorization using [Pollard's-rho algorithm](https://en.wikipedia.org/wiki/Pollard%27s_rho_algorithm). This is related to #70.


examples/rho_pollard.nim
```diff
@@ -0,0 +1,46 @@
...
+    return some(divisor)
+
+
+func somePoly(number: BigInt): BigInt =
+  return number*number+1.initBigInt
+
+
+proc main() =
+  const someNum = "44077431694086786329".initBigInt
+  let result = rhoPollard(someNum, somePoly)
```
konsumlamm, line 38:
```suggestion
  let result = pollardRho(someNum, somePolynomial)
```

examples/rho_pollard.nim
```diff
@@ -0,0 +1,46 @@
...
+  if result.isSome():
+    echo fmt"{result.get()} is a factor of {someNum}"
+    assert someNum mod result.get() == 0.initBigInt
```
konsumlamm, line 41:
```suggestion
  if result.isSome():
    let factor = result.get()
    echo fmt"{factor} is a factor of {someNum}"
    assert someNum mod factor == 0
```

examples/rho_pollard.nim
```diff
@@ -0,0 +1,46 @@
...
+  else:
+    echo fmt"could not find a factor of {someNum} using rho-Pollard algorithm"
```
konsumlamm, line 43:
```suggestion
    echo fmt"could not find a factor of {someNum}"
```
This doesn't mean that one can't find a factor using the algorithm, one might have to use a different polynomial or a different initial value.

examples/rho_pollard.nim
```diff
@@ -0,0 +1,46 @@
...
+    hare = initialValue
+    divisor = 1.initBigInt
+
+  while divisor == 1.initBigInt:
+    turtle = nextIterationMod(turtle)
+    hare = nextIterationMod(nextIterationMod(hare))
+    divisor = gcd(turtle-hare, n)
+
+  if divisor != n:
+    return some(divisor)
```
konsumlamm, line 29:
```suggestion
func pollardRho(
  n: BigInt,
  polynomial: proc(x: BigInt): BigInt {.noSideEffect.},
  initialValue: BigInt = 2.initBigInt): Option[BigInt] =
  ## Performs Pollard's rho algorithm to find a non-trivial factor of `n`.
  func polynomialMod(x: BigInt): BigInt =
    polynomial(x) mod n

  var
    turtle = initialValue
    hare = initialValue
    divisor = 1.initBigInt

  while divisor == 1:
    turtle = polynomialMod(turtle)
    hare = polynomialMod(polynomialMod(hare))
    divisor = gcd(turtle - hare, n)

  if divisor != n:
    some(divisor)
  else:
    none(BigInt)
```

examples/rho_pollard.nim
```diff
@@ -0,0 +1,46 @@
...
+
+
+func somePoly(number: BigInt): BigInt =
+  return number*number+1.initBigInt
```
konsumlamm, line 33:
```suggestion
func somePolynomial(x: BigInt): BigInt =
  x * x + 1.initBigInt
```

examples/rho_pollard.nim
vil02: Why do you insist on having the _else_ branch and explicitly returning the `none(BigInt)`?

examples/rho_pollard.nim
```diff
@@ -0,0 +1,46 @@
+## Rho-Pollard
+##
+## This file illustrates how to find a factor of an integer using the
+## [rho-Pollard algorithm](https://en.wikipedia.org/wiki/Pollard%27s_rho_algorithm).
```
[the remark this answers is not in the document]
vil02: Renaming is done in d785567d3c42aeba1fc7a2b8a2877ab50494d06b.

vil02:
@konsumlamm thanks for a review!

examples/rho_pollard.nim
konsumlamm: I think it's clearer to explicitly return `none(BigInt)` than to rely on it being the default value for `Option`s.

vil02: Please have a look at 145dc8858b6e24a64dc956d222db47e5a71c6a3b.

konsumlamm (approved):

dlesnoff (approved):

konsumlamm:
@narimiran can this be merged?

PR #130, opened 2023-06-29
chore: split workflows

vil02:
This PR is an attempt to fix the [`test.yaml`](https://github.com/nim-lang/bigints/blob/ca00f6da386af9ad7e3abf603c0201da6a014477/.github/workflows/test.yaml) workflow. It moves the `doc` part into a new file. Also I applied the `yamllint` suggestions (except the `line-length`).

The `test.yaml` part [_works on my side_](https://github.com/vil02/bigints/actions/runs/5414749807). I do not really have a way to verify the `doc.yaml` part.

#126 is related.

### DIFF
--- a/.github/workflows/doc.yaml
+++ b/.github/workflows/doc.yaml
@@ -1,0 +1,29 @@
+---
+name: docs
+
+# yamllint disable-line rule:truthy
+on:
+  push:
+    branches:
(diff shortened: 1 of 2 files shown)


dlesnoff:
Fine by me. Thanks!

Issue #123, opened 2023-01-12
Division causes assert failure


guidovranken:
```nim
import bigints

var a = initBigint("6277101735386680763835789123314955362437298222279840143829")
var b = initBigint("1461501637330902918203684832716283019655932313743")
var r = a.div(b)
```

Output:

```
Error: unhandled exception: /home/jhg/.nimble/pkgs/bigints-1.0.0/bigints.nim(652, 14) `q1 <= uint32.high`
```

Using latest repository version of `bigints`, Nim 1.6.10 on Linux x64.

pietroppeter:
Wow that's bad, thanks for reporting! I can reproduce on playground: https://play.nim-lang.org/#ix=4kYW

pietroppeter:
I checked going back to older commits and it seems to be a very old issue: already present in (Aug 2020, after merging a PR by Araq to make it work with nim devel) https://github.com/nim-lang/bigints/commit/8cb412f2302c3753999be395ae004ee836441277

going further in the past is hard since I could not find a nim version that would make bigints compile without additional changes. probably not worth it to go that way back.

pietroppeter:
doing a bit of debugging but need to stop now, will try to look into this later (if someone wants to pick this up, most welcome)

for ref this is how I annotated `unsignedDivRem` (where the bug occurs)

```nim
import sugar

# From Knuth and Python
func unsignedDivRem(q, r: var BigInt, n, d: BigInt) =
  var
    nn = n.limbs.len
    dn = d.limbs.len
  dump nn
  dump dn

  if n.isZero:
    q = zero
    r = zero
  elif nn < dn:
    # n < d
    q = zero
    r = n
  elif dn == 1:
    var x: uint32
    unsignedDivRem(q, x, n, d.limbs[0])
    r.limbs = @[x]
    r.isNegative = false
  else:
    assert nn >= dn and dn >= 2

    # normalize
    let ls = 32 - bits(d.limbs[d.limbs.high])
    r = d shl ls
    q = n shl ls
    if q.limbs.len > n.limbs.len or q.limbs[q.limbs.high] >= r.limbs[r.limbs.high]:
      q.limbs.add(0'u32)
      inc(nn)

    let k = nn - dn
    assert k >= 0
    var a: BigInt
    a.limbs.setLen(k)
    let wm1 = r.limbs[r.limbs.high]
    let wm2 = r.limbs[r.limbs.high-1]
    var ak = k

    var zhi = zero
    var z = zero
[...]

PR #133, opened 2023-07-23
Fix `shl`, `shr` & simplify `isZero`

konsumlamm:
Now `a: BigInt` is zero if and only if `a.limbs` is empty. This simplifies `isZero` and reduces special cases to consider when consuming a `BigInt`.

While checking that all functions uphold this invariant, I noticed that `shl` and `shr` were buggy:
* `shl` didn't check for zero, which allowed the user to construct `BigInt`s with arbitrarily many zeros in its limbs, breaking an invariant
* `shr` didn't check if the shift is too big, which could result in an exception

### DIFF
--- a/src/bigints.nim
+++ b/src/bigints.nim
@@ -9,7 +9,7 @@ type
     ## An arbitrary precision integer.
     # Invariants for `a: BigInt`:
     # * if `a` is non-zero: `a.limbs[a.limbs.high] != 0`
-    # * if `a` is zero: `a.limbs.len <= 1`
+    # * if `a` is zero: `a.limbs == @[]`
     limbs: seq[uint32]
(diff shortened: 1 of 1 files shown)


konsumlamm:
Division is making too much problems, perhaps I'll reconsider this once division has been rewritten.

dlesnoff:
If you have some failing checks, maybe this could be used for fixing the division algorithm. Do not hesitate to share these checks on the #123 issue.

konsumlamm:
The problems are that `unsignedDivRem` relies on 0 having a limb in a lot of places and trying to replace those cases manually is a pain. Some of these cases should go away once we have operations with `SomeInteger` operands, but eventually it should probably get rewritten.

PR #134, opened 2023-07-23
Fix `shl` and `shr`

konsumlamm:
`shl` could create invalid zero values, by shifting 0.

`shr` could raise an exception when the shift is too big.

### DIFF
--- a/src/bigints.nim
+++ b/src/bigints.nim
@@ -479,6 +479,9 @@ func `shl`*(x: BigInt, y: Natural): BigInt =
     assert a shl 1 == 48.initBigInt
     assert a shl 2 == 96.initBigInt
 
+  if x.isZero:
+    return x
+
   var carry = 0'u64
   let a = y div 32
   let b = uint32(y mod 32)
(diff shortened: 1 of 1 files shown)


src/bigints.nim
```diff
@@ -505,6 +508,8 @@ func `shr`*(x: BigInt, y: Natural): BigInt =
   let a = y div 32
   let b = uint32(y mod 32)
   let mask = (1'u32 shl b) - 1
+  if a >= x.limbs.len:
+    return zero
```
dlesnoff, line 512: I would put that before computation of b and mask.

dlesnoff (changes requested):
This is nicer, but I would avoid a few computations in this edge case.

src/bigints.nim
konsumlamm: I would expect the (C) compiler to automatically reorder that, but sure, I can do that.

konsumlamm:
@narimiran can this be merged?

PR #135, opened 2023-07-23
Improve `powmod`

konsumlamm:
* avoid redundant `modulo` call if `exponent >= 0`
* avoid shifting `exponent`

Unfortunately, I didn't notice any difference in performance, but this should reduce the number of allocations.

### DIFF
--- a/src/bigints.nim
+++ b/src/bigints.nim
@@ -1244,7 +1244,7 @@ func fastLog2*(a: BigInt): int =
   ## If `a` is zero, returns -1.
   if a.isZero:
     return -1
-  bitops.fastLog2(a.limbs[^1]) + 32*(a.limbs.high)
+  bitops.fastLog2(a.limbs[^1]) + 32 * (a.limbs.high)
 
 
 func invmod*(a, modulus: BigInt): BigInt =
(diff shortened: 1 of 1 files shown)


mratsim:
Here is a guide on how to implement fast `powmod` https://github.com/status-im/nim-stint/issues/126

The result is +/- 10% faster than GMP, without assembly: https://github.com/status-im/nim-stint/issues/126#issuecomment-1567181309

image

dlesnoff:
We trade one shr and a minus sign by a div, a shr and a mod?
@konsumlamm Have you compared the generated assembly?
Is there some optimizations with fastLog that I missed?
For which limb sizes did you benchmark it?
@mratsim is mainly proposing to switch to the clang compiler and perform some algorithmic optimizations too. Do you think it could be doable to switch to the clang compiler when clang is available on the user's computer?

konsumlamm:
@mratsim thanks for the link! I created a new issue for further optimizations (https://github.com/nim-lang/bigints/issues/137).

konsumlamm:
@dlesnoff @narimiran what do you think?

dlesnoff:
I am not convinced. The exponent being an int in your code, it would allocate at most a two-limb BigInt in my code (and for most of the uses, it will be a single limb).

Is the cost of memory allocation so huge that it is better to perform a few more computations?
> I didn't notice any difference in performance,

It does not seem so, on your computer at least. The number of allocations might impact on more constrained environments (embedded) but I do not think that we target embedded. At least, there are algorithmic changes to perform before micro-optimizations.

I would be interested to see either (or both):
– A reproducible benchmark showing the differences in performance and memory usage (not in the PR since that will be difficult but in the conversation).
– An overload of the procedure to give a separate interface for people having an int as exponent.

PR #138, opened 2023-09-24
feat: add Pollard's p-1 example

vil02:
This PR adds an example of factorization using [Pollard's p-1 algorithm](https://en.wikipedia.org/wiki/Pollard%27s_p_%E2%88%92_1_algorithm). This is related to #70 and similar to #129.

### DIFF
--- a/examples/pollard_p_minus_1.nim
+++ b/examples/pollard_p_minus_1.nim
@@ -10,14 +10,14 @@ import std/strformat
 
 func pollardPMinus1(
   n: BigInt,
-  searchLimit: Natural,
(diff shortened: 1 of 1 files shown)


konsumlamm:
Given that we already have an example for Pollard's rho algorithm, which looks pretty similar, and that Pollard's p-1 algorithm only works for certain numbers, I'm not sure we should add this.

dlesnoff:
I am okay with adding this example. It adds a little burden maintaining those examples, especially when the API is not perfectly stable, or the Nim version changes. Let's close #70 after this PR and focus on the library.

Issue #94, opened 2022-01-28
Compound assignment operators shouldn't use `template`s


konsumlamm:
Currently, e.g. `+=` is defined as
```nim
template `+=`*(a: var BigInt, b: BigInt) =
  a = a + b
```
This causes `a` to be evaluated twice, which can produce unexpected results, when it's an expression that can cause side effects. For example
```nim
proc lol(x: var BigInt): var BigInt =
  echo "hello ", x
  x

var a = 42.initBigInt
lol(a) += 1.initBigInt
```
prints `hello 42` twice, instead of just once.

The solution would be to just make it a `func` instead. The same applies to `-=` and `*=`.

konsumlamm:
As a further point, `+=` etc. in the standard library are also `proc`s/`func`s.

konsumlamm:
I tried to implement this, but changing `+=` breaks a test. This is possibly caused by https://github.com/nim-lang/Nim/issues/19464.

dlesnoff:
Could you tell us which test is broken by this change ?

konsumlamm:
> Could you tell us which test is broken by this change ?

The first "off by one in division" test case in `tests/tbugs.nim`. You can also check this for yourself by simply making `+=` a `func`.

dlesnoff:
This is indeed strange, it is off by one like in issue #5 (but one greater than expected result), and only for the static call. Otherwise I didn't encounter any other bugs. We might have a look at the changes of the PR #6 that should have fixed issue #5.

dlesnoff:
@konsumlamm Isn't this issue now just a matter of changing template to proc with a :%s/template ′(\c)=′/proc ′$1=′/g (Vim approximative sed regexp to change all template of compound assigments into procs).

konsumlamm:
That fixes the issue for 1.6, but not for 1.4. So there are several possibilities for how to proceed:
* drop 1.4 support
* accept that compound assignment operators are buggy for 1.4 (either by keeping them as `template`s or by changing them to `func`s too) - this should only be a temporary solution imo
* reimplement compound assignments to not copy the first argument (and possibly reimplement `+` as `result = a; result += b`) - I think this is the best option in the long run

PR #141, opened 2023-12-03
Test on i386

### LINKED ISSUE #48: Why is this library not expected to work on 32 bit?
The readme states this under [Current limitations and possible enhancements](https://github.com/nim-lang/bigints#current-limitations-and-possible-enhancements), but I don't see any obvious reason why this is the case. Could anyone elaborate on what exactly breaks on 32 bit platforms?

konsumlamm:
Test on i386 (a 32 bit arch) in the CI and update the versions tested.

Closes #48.

The CI job uses the prebuilt binaries from https://nim-lang.org/install.html (since most actions aren't available in a container). This means we can finally confirm that the library works on 32 bit architectures. I had to fix an example and the `rand` code (to work around https://github.com/nim-lang/Nim/issues/16360). I also added a test for `r.rand(x..x)`.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -14,19 +14,48 @@ jobs:
       - run: echo "commit message doesn't contain '[skip ci]'"
 
   test:
+    needs: before
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         nim-version:
           - '1.4.8'
-          - '1.6.12'
+          - '1.6.16'
+          - '2.0.0'
           - 'devel'
-    needs: before
     steps:
       - uses: actions/checkout@v4
       - uses: jiro4989/setup-nim-action@v1
         with:
           nim-version: ${{ matrix.nim-version }}
+      - run: nimble install -y
+      - run: nimble test
+      - run: nimble checkExamples
+
+  i386:
+    needs: before
+    runs-on: ubuntu-latest
+    container:
+      image: i386/ubuntu:bionic
+    strategy:
+      fail-fast: false
+      matrix:
+        nim-version:
+          - '1.4.8'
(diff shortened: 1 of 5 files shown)


.github/workflows/test.yaml
```diff
@@ -14,14 +14,16 @@ jobs:
       - run: echo "commit message doesn't contain '[skip ci]'"
 
   test:
+    needs: before
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         nim-version:
           - '1.4.8'
```
dlesnoff, line 23: We might want to drop support for 1.4.6.

tests/trandom.nim
```diff
@@ -14,9 +14,13 @@ block: # check uniformity
     doAssert(lo <= r)
     doAssert(r <= hi)
     total += r
-    let iBucket = (r-lo) div ((hi-lo) div initBigInt(nbuckets))
+    let iBucket = (r - lo) div ((hi - lo) div initBigInt(nbuckets))
     buckets[iBucket.toInt[:int]().get()] += 1
   for x in buckets:
-    doAssert(trials/nbuckets*0.5 < float(x))
-    doAssert(float(x) < trials/nbuckets*1.5)
+    doAssert(trials / nbuckets * 0.5 < float(x))
```
dlesnoff, line 20: I am unsure of the operator precedence here. If we multiply the numerator by .5, then the constant should be put in front of the expression. If we multiply the denominator by .5, then parenthesis should be added around the denominator.

.github/workflows/test.yaml
konsumlamm: I wouldn't mind that, 1.4 has quite some bugs (#59, #93, #94) and doesn't receive backports anymore.

PR #142, opened 2023-12-04
Unify `initBigInt` for `SomeInteger`s

konsumlamm:
Unify the various `initBigInt` versions that convert an integer to a `BigInt` into one `func`.

### DIFF
--- a/src/bigints.nim
+++ b/src/bigints.nim
@@ -37,39 +37,32 @@ func initBigInt*(vals: sink seq[uint32], isNegative = false): BigInt =
   result.isNegative = isNegative
   normalize(result)
 
-func initBigInt*[T: int8|int16|int32](val: T): BigInt =
-  if val < 0:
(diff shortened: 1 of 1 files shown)


pietroppeter:
FWIW I do prefer the original version and I am not able to see advantages of the unification.

konsumlamm:
Before, `initBigInt` for `int` and `uint` were templates, so it wasn't possible to e.g. pass them to higher order functions. Additionally, now you can specify the integer type in all cases. I think it's more consistent to have one function for all integer types than have some with a generic argument and some without (and some templates). This also reduced the number of cases (there is no extra case needed for `int` and `uint` anymore).

pietroppeter:
Ok, thanks for the explanation

dlesnoff:
I had the same impression as @pietroppeter and I also noticed that it greatly simplifies the documentation with just what we want in the public API.
Are you just sure that it was not purposefully chosen to use templates there in the first place?

dlesnoff (approved):
Thanks.

PR #152, opened 2026-09-21
Fix shr on negative numbers and quotient digit overflow in div/mod

Balans097:
This PR fixes two independent bugs in `src/bigints.nim`, one per commit. Each has a regression test in `tests/tbugs.nim`.

### 1. `shr` returns 0 for negative numbers when all limbs are shifted out

`shr` is an arithmetic shift (floor division by 2^y, 2's complement semantics). When the shift amount removes every limb, it returned `zero` unconditionally. For a negative operand the correct result is `-1`:

```nim
doAssert (-1).initBigInt shr 32 == (-1).initBigInt   # returned 0
doAssert (-2).initBigInt shr 33 == (-1).initBigInt   # returned 0
```

Shifts that keep at least one limb already round correctly (the `underflow` branch), so only this early-return path was affected.

### 2. Quotient digit estimate can reach 2^32 in `unsignedDivRem`

In Knuth's Algorithm D, step D3, when the leading limb of the current dividend window equals the leading divisor limb (`vtop == wm1`), `vv div wm1` is >= 2^32 and has to be clamped to b-1, with `rhat = vv - qhat*wm1`. This clamp was missing, so:

- in default builds `assert q1 <= uint32.high` fires (`AssertionDefect`);
- with assertions disabled (e.g. `-d:danger`) the oversized digit is truncated and the result is silently wrong.

The refinement loop is now only run while `rhat < b`, as in step D3.

Minimal reproducer:

```nim
let n = "730750818495310275721298203196880063083711561727".initBigInt
let d = "39614081247908796766359650303".initBigInt
doAssert n div d == "18446744073709551615".initBigInt
doAssert n mod d == "696856065384891421382672382".initBigInt
```

The new regression tests fail before and pass after the change; the existing test suite passes.


narimiran:
Can you rebase and resolve conflicts? (One of the fixes might not be needed anymore, as another PR was merged)

